PR #69, opened 2016-01-17
Diagonal laser sprites

Cruix:
Adds support for 8-way directional sprites on projectiles. Does not change the path the projectiles take, only their sprite. Projectiles with icon states that have only 4 directional states appear to still work normally. Also adds 8-state directional sprites for the following icon states in projectiles.dmi:
laser
heavylaser
xray
u_laser
bluelaser
omnilaser
ice_2
declone


TehFlaminTaco:
I recommend swapping that out, and keeping all lasers dirs set to 1, but setting their transform to a turn(transform, atan2(current.x-x, current.y-y))

Makes them point directly at their target.

Cruix:
That would also add slightly fewer resources for people to download, I think, and it would make it more clear exactly where the projectile is going, but it some cases it would cause the laser to appear to be moving sideways. I'll keep this the way it is, but you can make a new pull request and let the Head Coders decide if you want.

Alek2ander:
There's also this 
https://github.com/tgstation/-tg-station/pull/14092
https://github.com/tgstation/-tg-station/pull/13458
https://github.com/tgstation/-tg-station/pull/13227
https://github.com/tgstation/-tg-station/pull/10818

...except it probably doesn't work.

Hydre:
69 lol

george99g:
How is this a fix?

Issue #49, opened 2016-01-16
Wizard Spells


george99g:
opened by Ben · 14 days ago
What is the nature of the issue: Read the screenshot
What steps must be taken to reproduce this error:
http://imgur.com/yDf7RzT
Any additional information? : please fix posthaste!

BadminBen:
Was fixed so closing

PR #72, opened 2016-01-18
Fixes issue #49

Super3222:
Fixes issue #49, the issue where refunding a spell with efficient or a level higher will allow you to buy it at that same level at the cost of 1.


AsV9:
Looks good.

Issue #37, opened 2016-01-16
Changlings


george99g:
opened by Ben · 2 months ago
What is the nature of the issue:Transformation Sting not working or working very rarely
What steps must be taken to reproduce this error:
Become changling get transformation sting
Attempt to sting someone
Fail
Any additional information? :

X-TheDark:
So far, seems that changing [tiny_prick.dm](https://github.com/yogstation13/yogstation/blob/c8b5bf85b54c7c0adb059925829a6246fd115999/code/game/gamemodes/changeling/powers/tiny_prick.dm) makes the Transformation Sting work.
Line 68 from:

```
var/datum/dna/selected_dna = null
```

To:

```
var/datum/changelingprofile/selected_dna = null
```

as well as line 90 to:

```
add_logs(user, target, "stung", "transformation sting", " new identity is [selected_dna.dna.real_name]")
```

and line 91:

```
var/datum/dna/NewDNA = selected_dna.dna
```

Xantam-13:
@X-TheDark - Perhaps open a pull request with this fix? :)

X-TheDark:
As soon as I figure out why Dream Daemon keeps crashing after 3 minutes of starting the game (right at the moment of the "Communications intercepted" thing), I'll get right on that.

X-TheDark:
Ok, so far I was able to make it work, it changes DNA data and doesn't break the game or anything, but it doesn't update the appearence of the target correctly (gender/skin color/eye color/hair/etc, it uses some default settings, such as Bald, Caucasian1 skin color, etc).

I'm going to see if I can make the sting also change the appearence of the target to the correct one in the nearest few hours.

PR #75, opened 2016-01-18
Transformation Sting fix (Issue #37)

X-TheDark:
Fixes issue #37.
Got Transformation Sting to work but gave up on making monkeys being transformed properly after like 5 hours of trying. It does work with lizardpeople/plantpeople/androids (no mental strength to look up their proper names after all the work).

Problems with this particular solution:
1. Incredibly hacky, since I basically use additional vars in changelingprofile to save and later set appearance data (even though it SHOULD be set after calling hardset_dna, but that doesn't seem to be the case).
2. Monkeys don't get the correct appearance because, basically, any call to updating deletes the old mob and after trying about 3 solutions, I just gave up on making monkeys work (I can only hope updateappearance() accounts for changeling monkeys).
3. Doesn't change the naked sprite (if you were wearing a bra, you will be wearing a bra even if turned into male).

In essence, hardset_dna() just doesn't seem to work properly. The UI is passed, but it doesn't get processed, which causes appearance to not get updated based on DNA and I have to resort to saving/setting this manually.

I'll leave it up to the coders whether or not to merge this. But hey, at least you can turn the captain into a lizardperson and kek for days.

### DIFF
--- a/code/game/gamemodes/changeling/changeling.dm
+++ b/code/game/gamemodes/changeling/changeling.dm
@@ -365,6 +365,14 @@ var/list/slot2type = list("head" = /obj/item/clothing/head/changeling, "wear_mas
 	prof.dna = new_dna
 	prof.name = H.real_name
 	prof.protected = protect
+	prof.gender = H.gender
+	prof.skin_tone = H.skin_tone
+	prof.hair_color = H.hair_color
+	prof.hair_style = H.hair_style
+	prof.facial_hair_color = H.facial_hair_color
+	prof.facial_hair_style = H.facial_hair_style
+	prof.eye_color = H.eye_color
+	prof.features = H.features
 
 	var/list/slots = list("head", "wear_mask", "back", "wear_suit", "w_uniform", "shoes", "belt", "gloves", "glasses", "ears", "wear_id", "s_store")
 	for(var/slot in slots)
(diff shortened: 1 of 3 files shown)


code/game/gamemodes/changeling/changeling.dm
```diff
@@ -460,6 +468,16 @@ var/list/slot2type = list("head" = /obj/item/clothing/head/changeling, "wear_mas
 	var/list/exists_list = list()
 	var/list/item_color_list = list()
 	var/list/item_state_list = list()
+	
+	//Xthedark : Human/Etc appearence data, seems we'll have to save/set this manually
```
TehFlaminTaco, line 472: You really don't need to add your name to your comments, that's bad practice.

You're supposed to use comments to angrily yell at the code.

george99g, line 472: Why would you add your name to the comment?
The information of who added it is all recorded by git.
No need to clutter it with names. This isn't comment-based collaboration.

X-TheDark, line 472: Damn, sorry. Force of habit. Should this be closed and recommited without the comment?

AsV9:
Not a major issue, I'll just merge this and you'll get it next time.

code/game/gamemodes/changeling/changeling.dm
AsV9, line 472: Not a major issue, I'll just merge this and you'll get it next time.

Issue #58, opened 2016-01-16
Changeling ghost bug.


ghost:
Step 1: Monkey or lesser form.

Step 2: Automatically ghost then get cremated because of it.

Also happens with genetic monkey syringes. Seems to happen if they're moving at the same time as they transform.

Should also mention it's really fucking hilariously abusable by xenobiology because if you click the headslug with a setinence potion and the ling's ghosted you get a pet headslug that can burst into a pet changeling, as you can with the gold slime thing.

Super3222:
This was fixed. @george99g <3

george99g:
Closing this then.

PR #80, opened 2016-01-18
Monkeyize fix

Cruix:
Fixes issue #58. Changelings using lesser form and humans using human-to-monkey DNA injectors no longer ghost if they try to move while transforming.


Cruix:
Well, something went very wrong here.

Issue #79, opened 2016-01-18
Make beepsky's lights alternate between red and blue, a la police car.


ghost:
BEEP BEEP MOTHERFUCKER THE POLICE IS ON THE WAY.

shrimptk:
This would look really cool with Taco's new lighting system, but if it were only active during arrests (e.g When beepers goes in for the kill) it would be kinda annoying to have it rapidly toggle on and off, however if it could be activated if someone called beepers using a PDA control and he started blaring lights and sirens on his way to the scene that would be fucking rad as shit.

TehFlaminTaco:
Resolved.

george99g:
Implemented by @Cruix with #84

PR #84, opened 2016-01-19
Justice Beepsky

Cruix:
Makes Beepsky (and only beepsky, not normal securitrons) flash red and blue whenever he is arresting someone or being summoned to a location. While flashing, he gains a sprite that makes his eyes flash and his light color changes between blue and red.

Response to issue #79


TehFlaminTaco:
Looks awesome, but should make it change every five ticks, so the lighting engine can update along side it.

code/game/machinery/bots/secbot.dm
```diff
@@ -33,6 +33,21 @@
...
+
+/obj/machinery/bot/secbot/beepsky/bot_process()
+	..()
+	if(on && (mode == BOT_PREP_ARREST || mode == BOT_HUNT || mode == BOT_ARREST || mode == BOT_SUMMON))
+		icon_state = "secbot_flashing"
+		if(flashing_timer % 2 == 0)
+			SetLuminosity(luminosity, 1, 0, 0)//red
+		else
+			SetLuminosity(luminosity, 0, 0, 1)//blue
+		flashing_timer++
```
george99g: Why increment and then use the remainder from the division?
Wouldn't the eventual overflow cause it to return negative values?
Why not just use a bool and toggle it?

DavidEdwards: It would be do-able if the var was capped. if(flashing_timer == 2)
flashing_timer = 0.

ghost:
its happening

Cruix:
flashing_timer was originally going to also be used to tell how long Beepsky had been flashing for, so it could turn off if it went on for too long, but I never implemented that failsafe.

I'm pretty sure it would still work if the variable overflowed. I'm not sure if DM's mod operator returns the real mod or the remainder, but since both 1 and -1 count as true for the "if" statement, it will work in either case.

That being said, I'll probably change it to a bool like you suggested in my next commit.

Cruix:
Updating the lights every 5 ticks works great, it cycles much faster now. thanks Taco.

TehFlaminTaco:
Code seems good, I'd say ready to merge.

Issue #48, opened 2016-01-16
Bomb Collars


george99g:
opened by Ben · 24 days ago
What is the nature of the issue: Bomb Collars being stuck in a mobs hand
What steps must be taken to reproduce this error:
Bomb collar was placed on a human
Human got turned into a monkey to remove collar and turned back into a human
As a human they picked up the bomb collar now lying on the floor and it got stuck in their hand.
Any additional information? : Will test in on test server

Super3222:
I've tested this on my own server, and it turns out the bug is real. The problem is that there is no other way to get it off your hand unless you put it on your own head. I believe this is because monkeyizing forces everything on them to fall to the ground, but doesn't change the bomb collars NODROP flag. So when you pick it up again, you cannot remove it from your hand because of that flag.

Super3222:
after much spaghetti, I've come up with a potential patch for this. I've been testing all day and can say as a witness the bug exists.

PR #90, opened 2016-01-19
Fix for Issue #48. Bomb Collar Bug.

Super3222:
Fixes issue #48 
The bug was where if you have a bomb collar locked to your head, you monkey out of it, become a human/or stay a monkey, and than decide to pick it up you can't drop it what so ever.

After a bit of spaghetti I was able to fix this.


Super3222:
Something went wrong

PR #103, opened 2016-01-20
Fixed some bot interface buttons being usable when they should not be.

### LINKED ISSUE #93: You can change some bot settings from any distance
I noticed while I was messing with Beepsky that many bot Topic options don't properly check for distance and access when you try to alter them. Options that are built in to the bot class, such as toggling power and toggling patrol mode, do properly check, but most of the options added in the more specific bots (such as toggling a secbot's detain mode) don't. I'll fix it after PR #84 is pulled or rejected, unless someone tells me that there is a reason for this.

Cruix:
Fixes #93. Fixed some bot interface buttons being usable when they should not be. Also made Beepsky's siren non-toggleable unless Beepsky is unlocked. Added some explicit "return 0"s in canUseTopic() proc where they were just "return"s.

### DIFF
--- a/code/game/machinery/bots/cleanbot.dm
+++ b/code/game/machinery/bots/cleanbot.dm
@@ -97,8 +97,8 @@ text("<A href='?src=\ref[src];power=1'>[on ? "On" : "Off"]</A>"))
 	return
 
 /obj/machinery/bot/cleanbot/Topic(href, href_list)
-
-	..()
+	if(..())
+		return 1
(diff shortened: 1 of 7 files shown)


TehFlaminTaco:
Does not compile, please fix this pull.

george99g:
> Does not compile, please fix this pull.

It does compile.
Travis just fucked up getting the compiler going and failed it.

george99g:
Seems fine.

Issue #104, opened 2016-01-20
Foam dart bug


00aquir:
Foam darts disappear if they hit Beepsky.

Cruix:
It's not just Beepsky or secbots, it's all bots.

Cruix:
It seems that bullets that hit bots like Beepsky only get their bullet_act() proc called if the bullet does brute or burn damage. Foam darts do 0 OXY damage on hit, so the proc that makes them fall to the ground is never called.

george99g:
Fixed in #105

PR #105, opened 2016-01-20
Fixed foam darts disappearing when they hit Beepsky

Cruix:
Fixes issue #104. Made it so that the base class's bullet_act() is always called when a bot is hit by a projectile. This means that foam force darts and other reusable ammo fired at bots no longer disappears.  Also made Mulebots process damaging projectiles properly again, since they depended on the old code.


george99g:
Code seems fine.

PR #138, opened 2016-01-25
Makes changelings immune to the zombie virus

### LINKED ISSUE #135: Make lings zombie-immune
Because it dicks with transforms, lings can't resist out with pancea or transform to avoid it and a random event that can instantly shit on everything is bad.

Please make lings zombie-immune, and naturally secrete the antibodies for zombie virus.

Cruix:
Fixes #135.


Xantam-13:
I'd actually say, perhaps that could be something they need to spend points on to become immune.

PR #141, opened 2016-01-25
Coloured Lighting Soft Revert

TehFlaminTaco:
Reverts the coloured lighting as it is until Byond fixes their shit, all
the code for coloured lights is still in place, but the main workload is
restored to it's alpha-only state. Very mild proformance boost to be
expected, but don't expect anything noticable. Fixes bugged clients.


george99g:
Seems good!

Issue #155, opened 2016-01-26
Add "Welcome to Lunar Industries" as a new lobby theme


ParagonPrime:
(https://youtu.be/_lAfMT5FIZE)

So, the theme song of Moon is hauntingly appropriate for an SS13 theme song- it just sounds fitting, never mind the fitting plot of it. The length is relatively good (I even tested it out) for the pre-game sequence as well, and it just creates some nice haunting atmosphere.

Also, we kinda need a bit more variety in our lobby music, so there you go.

Implement pl0x!

george99g:
Do keep in mind that the size of the resources everyone needs to download increases with every song we add in.

ParagonPrime:
Then perhaps we could remove that third lobby theme that almost never plays with this one? And in any case, as the community voted in favour of adding new music... why not this one?

X-TheDark:
This may need to be edited to be shorter before being added.

george99g:
> Then perhaps we could remove that third lobby theme that almost never plays with this one?

Regarding that, it actually never ever played. #157

george99g:
A council vote regarding this is now underway.

george99g:
Council vote passed.
A trimmed version was added to #157 and merged.
Closing this.

PR #157, opened 2016-01-26
Adds title3.ogg to the lobby music rotation and replaces it with #155's suggestion.

george99g:
Why was this not included before?
Also, untested since I wrote this on Linux.


AsV9:
> "Needs testing"

Wot

DavidEdwards:
He didn't compile and run it because he was writing it on Linux. So he
can't guarantee it will work / compile. Testing tag is the best thing to do.

PR #158, opened 2016-01-26
Fixes Some Zombie Messages

Cruix:
Fixes zombies saying "This message will be turned into urrs." in deadchat and swaps the local and self messages in zombie/attack_paw()


ghost:
this message will be turned into urrs

PR #162, opened 2016-01-27
Fix for syndicate borgs being unable to use grenade launcher

X-TheDark:
Recent escapades as a syndicate borg have indicated problems with their (in)ability to use the internal grenade launcher, this has been fixed now.

Changed fire pin to standard.


DavidEdwards:
Is this going to affect anything else? It doesn't appear to be a borg specific weapon.

X-TheDark:
Not to worry! This actually **is** a cyborg specific weapon that is only added using the syndicate module. There does not exist any other way to obtain this weapon in-game.

If you make a repo search using "grenadelauncher/cyborg", you will find only 2 places it is used - one for definition (which I changed) and one for robot_modules.dm, where the Syndicate robot module is defined.

DavidEdwards:
Alright, then I'm game.

PR #176, opened 2016-01-29
Syndicate borg cost decreased to /tg/'s 80

X-TheDark:
Because screw paying 120 TC, which is 20 TC short of a Mauler for one. No wonder noone buys the damn thing, you can barely afford it at all.

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -233,7 +233,7 @@ var/list/uplink_items = list()
 	name = "Syndicate Cyborg"
 	desc = "A cyborg designed and programmed for systematic extermination of non-Syndicate personnel."
 	item = /obj/item/weapon/antag_spawner/borg_tele
-	cost = 120
+	cost = 80
 	gamemodes = list(/datum/game_mode/nuclear)
 	surplus = 0
 


AdamElTablawy:
👌👀👌👀👌👀👌👀👌👀 good shit go౦ԁ sHit👌 thats ✔ some good👌👌shit right👌👌there👌👌👌 right✔there ✔✔if i do ƽaү so my self 💯 i say so 💯 thats what im talking about right there right there (chorus: ʳᶦᵍʰᵗ ᵗʰᵉʳᵉ) mMMMMᎷМ💯 👌👌 👌НO0ОଠOOOOOОଠଠOoooᵒᵒᵒᵒᵒᵒᵒᵒᵒ👌 👌👌 👌 💯 👌 👀 👀 👀 👌👌Good shit

X-TheDark:
The suggestion to decrease cost to 60 TC stemmed from belief that it costs 80 TC which is obviously not the case. Further decrease to 60 TC still stands as a possibility.

DavidEdwards:
If TG did it, then it is probably "OK". Some people expressed concern at
making the borg less expensive.

X-TheDark:
This (80 TC cost) was assumed to be the baseline for comparison (from which the decrease to 60 TC was proposed). But since right now the borg is much more expensive than assumed at first (120 TC), it's probably a good idea to first decrease the cost to reasonable (read, /tg/) levels.

Plus, with the challenge mode requiring less players, we might actually see some consistent use (hopefully), not to mention it's grenades were fixed, so it should at least become a viable non-bug-ridden option.

DavidEdwards:
In my opinion, this all checks out. We can always raise it again if it is too far.

Issue #185, opened 2016-01-30
Remove Glowshrooms


Cephalopoet:
They're laggy, they multiply and add nothing to the game.

Time-Green:
They add pretty lights!

shrimptk:
+1 Either optimize them, or drop 'em

Xantam-13:
Drop 'em. They are useless.

Cruix:
But how is botany supposed to keep running when the lights go out? Glowshrooms are far more thematic than flashlights.

And as the only source of radium that can be grown, they are the only sustainable way botany can get extra-species mutations without the help of chemistry.

shrimptk:
@05rhardy The king has spoken, these powergamey shits need to get the fuck out

X-TheDark:
One word - Shadowlings. Other than that, they do more harm than good due to causing some mad lag.

Cruix:
Are we removing flares too?

Shadowlings can kill glowshrooms, they just slow them down. They can't do shit to flares.

Cruix:
Is this resolved by #187, or are we still thinking about removing glowshrooms?

DavidEdwards:
For the time being, this issue will be closed.

Cruix made some nerfs to Glowshrooms. If there is another uproar we can create a new issue.

PR #187, opened 2016-01-30
Major Glowshroom nerfs, some bugfixes.

Cruix:
Response to issue #185.
Significantly nerfs glowshroom spread in a desperate attempt to keep glowshrooms in the game. Also fixes some long-standing bugs where stats would not have the proper effects on later generations of glowshrooms.


george99g:
Seems good.

Issue #167, opened 2016-01-27
Add a tooltip to sleepers.


ghost:
Mousing over the sleeper buttons should provide a description of what they do, treat, and their overdose levels.

To prevent shitty med from pressing every button and killing a clown with morphine OD.

X-TheDark:
Sounds like something that's pretty useful.
Perhaps we could also add some sort of limiter that prevents people from OD'ing patients on drugs using the sleeper (unless emag'ed?).

ghost:
No, because forcing antags to not be able to OD people in sleepers unless they have a traitor-only item is shit.

Emagging a sleeper should make it automatically inject the maximum amount of chemicals into whoever's in it, until screwdrivered.

PR #201, opened 2016-01-31
ToolTip feature

Super3222:
Adds a tooltip feature to Yog. Courtesy of #167. 

Q: What is a tooltip?
A: It's basically a small menu that comes down when you hover your mouse over one of the alert icons. (ex: weightless, no oxygen, too much pressure, etc.), it provides the description of the alert as well. So you no longer have to shift click it.

Q. W-why all the edits on the skin.dmf Super? You trying to destroy the game?
A. It took some time to actually get tooltips on there. At first I was confused, but than I realized something. The tooltips were BEHIND the actual main screen, so I had to rearrange them to be the very first.


Xantam-13:
Sorry I may have missed the point or context. What exactly do they show? Can you give me an example of when I might need this or what it'd say when I hovered over? 
I only ask because I don't fully understand

Super3222:
tooltip

@05rhardy, It'll look like this when your mouse is over the alert.

Xantam-13:
Great. Is there a delay, or does the tip disappear immediately? If there's a delay, for admins mainly, would this interfere with pressing buttons? Obviously if there is no delay this question is irrelevant. 

Are there any known bugs? Have you tested it?

Super3222:
The tip disappears immediately when you take your mouse off of the alert it's over.

I also did test & search for bugs, though you never know what players might discover. I also had a look at /tg/s git to see if there's any current bugs that haven't been fixed. Basically everything's fine.

oisin100:
Looks epic! Would be pretty cool if it would work for other things such as sleepers and chemistry too!

DavidEdwards:
@Super3222 I need you to re-create this PR - so that it will base on the "dev" branch. That way I can easily test it on the dev branch, rather than having it merge onto master.

Super3222:
Will do.

Issue #170, opened 2016-01-28
Poster transformation bug


00aquir:
The most terrible bug ever for any spessman.. Once you abort (or someone (like security) makes you abort) putting a Nanotrasen poster onto a wall (you walk away while applying it, etc.), it changes and looks like a contraband item both in its color and description.

![ss13-posterbug](https://cloud.githubusercontent.com/assets/16806055/12641613/bd9b706a-c5b0-11e5-84de-ae9d7016a6bf.png)

george99g:
While someone's at it, could they also fix the "it's" to "its" on the Nanotrasen one?

00aquir:
This bug has changed now. After the fix, the Nanotrasen poster only turns into a red one after someone removes it from the wall with a wirecutter.

PR #209, opened 2016-02-01
Bug Fix + Hotkey Addition

ghost:
I'm new to GitHub and I was told by 05rHardy to make a pull request with my code. I'm not sure if this is the correct way to do things. I've added the ability to press z hotkey for reagent containers to change their transfer amounts and I've fixed issue #170, the poster transformation bug.

### DIFF
--- a/code/game/objects/effects/decals/contraband.dm
+++ b/code/game/objects/effects/decals/contraband.dm
@@ -79,7 +79,7 @@
 	anchored = 1
 	var/serial_number	//Will hold the value of src.loc if nobody initialises it
 	var/ruined = 0
-	var/subtype
+	var/subtype = 0
 
 /obj/structure/sign/poster/New(serial,subtype)
 	serial_number = serial
@@ -358,9 +358,10 @@
 			return
 
 	user << "<span class='notice'>You start placing the poster on the wall...</span>"	//Looks like it's uncluttered enough. Place the poster.
-
+	var/legit = istype(P, /obj/item/weapon/contraband/poster/legit)
 	//declaring D because otherwise if P gets 'deconstructed' we lose our reference to P.resulting_poster
 	var/obj/structure/sign/poster/D = P.resulting_poster
+	qdel(P)	//delete it now to cut down on sanity checks afterwards. Agouri's code supports rerolling it anyway
 
 	var/temp_loc = user.loc
 	flick("poster_being_set",D)
@@ -372,11 +373,10 @@
 
 	if(istype(src,/turf/simulated/wall) && user && user.loc == temp_loc)	//Let's check if everything is still there
 		user << "<span class='notice'>You place the poster!</span>"
-	else if(istype(P, /obj/item/weapon/contraband/poster/legit))
+	else if(legit)
 		D.roll_and_drop_legit(temp_loc)
 	else
 		D.roll_and_drop(temp_loc)
-	qdel(P)	//delete it now to cut down on sanity checks afterwards. Agouri's code supports rerolling it anyway //moved by Groxic to fix bug and check type
 	return
 
 //Putting non-contraband posters here because everything else here is related to posters anyway. -JS
--- a/code/game/objects/effects/decals/contraband.dm
+++ b/code/game/objects/effects/decals/contraband.dm
@@ -79,7 +79,7 @@
 	anchored = 1
 	var/serial_number	//Will hold the value of src.loc if nobody initialises it
 	var/ruined = 0
-	var/subtype = 0
+	var/subtype
 
 /obj/structure/sign/poster/New(serial,subtype)
 	serial_number = serial
@@ -337,7 +337,11 @@
 	P.resulting_poster = src
 	P.loc = location
 	loc = P
-
+/obj/structure/sign/poster/proc/roll_and_drop_legit(turf/location)
+	var/obj/item/weapon/contraband/poster/legit/P = new(src, serial_number)
+	P.resulting_poster = src
+	P.loc = location
+	loc = P
 
 //seperated to reduce code duplication. Moved here for ease of reference and to unclutter r_wall/attackby()
 /turf/simulated/wall/proc/place_poster(obj/item/weapon/contraband/poster/P, mob/user)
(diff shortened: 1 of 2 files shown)


Cruix:
Github is an enormous pain sometimes but it's much better than the alternative of manually merging everything.

Your pull request has made no changes, which you can see under the "files changed" tab. I suspect this is because you either didn't commit or didn't push. Commit your changes and push them to your repository, and this pull request should automatically update.

You are also using the master branch of your own fork, which isn't a problem for us, but could easily turn out to be a pain for you. I made the same mistake when I made my first pull request. Depending on what tool you are using to interact with Github, I may be able to help you fix it after we fix this pull request.

ghost:
I tried to fix it by making a new branch and them committing/pushing to that one and now I don't have the right changes that I wanted to push.

Cruix:
Are the changes still on your master branch?

ghost:
I don't think so.

ghost:
I had 8 uncommitted changes, now 6.

Cruix:
Looks like the only commit pushed to your repository is a merge with the Yogstation dev branch. Make sure you are staging all of your changes before you commit.

ghost:
I think I found the problem. I merged something or other with mine and I lost the changes I made. I'm fixing it now and I will commit it soon.
I'm using the GitHub desktop application thing.

ghost:
I made a new commit with the two files I changed and nothing is happening.

Cruix:
Did you push the commit?

If you did not, when you do, try to push it to a new be branch labeled "bugfix + hotkey addition" or something similar

ghost:
Sorry, but, how do I push?

ghost:
I've undone my last commit. You want me to make a new branch, commit then, and then what?

Cruix:
Push should be a button near commit, it is what actually sends the commit to your repository. If you wait a few hours I can check out the Github desktop app and find it.

ghost:
github

Cruix:
Try the sync button in the upper right

ghost:
It worked. Thank you so much.

Cruix:
Nice job getting Git working. I'll have some feedback on your code in a few hours.

ghost:
Thanks.

code/game/objects/effects/decals/contraband.dm
```diff
@@ -369,8 +372,11 @@
 
 	if(istype(src,/turf/simulated/wall) && user && user.loc == temp_loc)	//Let's check if everything is still there
 		user << "<span class='notice'>You place the poster!</span>"
+	else if(istype(P, /obj/item/weapon/contraband/poster/legit))
+		D.roll_and_drop_legit(temp_loc)
 	else
 		D.roll_and_drop(temp_loc)
+	qdel(P)	//delete it now to cut down on sanity checks afterwards. Agouri's code supports rerolling it anyway //moved by Groxic to fix bug and check type
```
george99g: Don't leave your name in comments, please.
Git already manages name-tracking.

Cruix:
Feedback:
First off, you should not make one pull request that adds a new feature and fixes an unrelated bug. You should make two separate branches and two separate pull requests. It may seem like a lot of trouble, but it helps if the repository owners want the bugfix but not the feature, or vice-versa.

I don't know why you changed line 82 in contraband.dm, but you should undo that. It won't change much, but there are some very subtle differences between the value 0 and the value null (which is what it is automatically assigned if you give it no default value).

The new roll_and_drop_legit() proc is good.

In the place_poster() proc, you did fix the bug. However, you moved the qdel from before the sleep() call to after it, so that you could check the type when you needed to. This leaves the poster in the hand of the person placing it, allowing them to place it multiple times before the first finishes unrolling. If the person does not move, only one of the posters is created on the wall, but if they do move, then two posters are dropped, which is a fairly serious duplication exploit. As in, if the clown gets one poster, soon the whole station will be covered in posters.

You can fix this by moving the qdel back to where it was, checking the type of P before the sleep() call, and storing the result in a variable that you look at when you need to.

In your reagent container code, you overrode the attack_self() proc twice, which you should almost never do. Furthermore, the only thing that the second attack_self() does is call the first one with the "..()" operator. You should remove the second attack_self() proc. (If you have experience in other object-oriented languages, you probably thought you were overriding two different overloaded procs, since they have different argument lists. However, DM does not care about the number or type of arguments when defining or calling a proc. Combined with DM's unorthadox mechanic of being able to override the same proc more than once, this can easily cause confusion.)

Because you used the attack_self() proc, people will be able to set the transfer amount not only with the hotkey, but also by selecting the reagent container in their active hand and clicking on it. This is fine, but you should make it more clear in your pull request - you emphasized the fact that it can be done with the hotkey, and did not mention that you could also do it by using the mouse.

In the first attack_self() proc, you clearly copy-pasted all the code from the set_APTFT() verb. While this worked, you should try to avoid copy-pasting code whenever possible, so that if the code is ever updated, you only have to update it in one place. Even in situations like this, where the copied and original code are right next to each other, you should still follow this convention.

You can fix this by just calling set_APTFT() in the attack_self() proc. Normally you would also have to add a range check, since the "set src in range(0)" line would not work when the verb is called by a proc like attack_self(), but because you can only call attack_self() when you are holding an item in your hand, there is no need for a range check. (With the exception of people using Telekenesis. If you care about people with TK being able to set the transfer amount of a beaker/bucket/etc. from long range, then add a range check or a contents check.)

Do not be discouraged. Though I described these issues in a very wordy way, these changes are very small, and for the most part your code does work as intended. By far the hardest part of coding is starting out, and the vast majority of people do not ever get to the point where they can even make a pull request.

ghost:
Thank you. Do I make the changes and then commit again? I didn't realize verbs could be called within procs. I couldn't find the original code for using hotkeys and I only found attack_self().

Cruix:
Yes, if you make changes, commit, and sync again this pull request should auto-update.

ghost:
Done.

ghost:
Also, I have another GitHub account, GRUBIK. Do you know why things, such as my merges/commits are going through that account?

Cruix:
I have no idea, sorry.

Cruix:
Code looks good, excellent work.

Issue #197, opened 2016-01-31
Remove confirmation window from destructive analyzer.


krgii:
Yes, I'm sure I want to destroy this item. I wouldn't have pressed destroy if I didn't want to.

Super3222:
I think the confirmation is there so people don't deconstruct something that's pointless towards research levels and can actually use. Considering deconstruction is permanent and you can't get it back unless it's re producible.

Maybe there could be a safety lock that you can toggle using the RD console with the destructive analyzer which goes ahead and deconstructs what's in it without asking you to confirm.

oisin100:
Maybe it should only prompt you when the item your deconstructing wont give you any research levels at all...

george99g:
> Maybe it should only prompt you when the item your deconstructing wont give you any research levels at all...

This would be an immense help to speed up R&D. The main part of it is checking if the item you're deconstructing will be any useful or not. With it prompting you and telling you when it's not, it'll take a lot of the work out.

oisin100:
Ill see what i can do about it then.

X-TheDark:
Done and closed.

PR #210, opened 2016-02-01
Change Destructive Analyzer to only prompt when item will be wasted

oisin100:
Title says it all.

Suggested at #197


DavidEdwards:
Excellent PR. Should speed things up!

Issue #195, opened 2016-01-31
Noslip shoes or Operation : Cancel Lubrication


X-TheDark:
So, I'm sure we can all agree that Space Lube is kind of ridiculous, but...how do we fix it?
Current proposals:
1. Make syndicate no-slip shoes not slip on lube;
2. New super-stik shoes that are more expensive but don't slip on lube;

Both of these options will further need to be discussed in context of their availability - Traitor and/or Nuke Ops.

Perhaps you have other ideas on how to make lube a bit less ridiculous? Or maybe it's perfectly balanced? Thoughts?

EDIT : Or maybe just change the operative Chem Sprayer to spray lube or something.

krgii:
Make magboots noslip too while active.

X-TheDark:
Done and merged.

PR #213, opened 2016-02-02
No-slip shoes and active magboots don't slip on anything

X-TheDark:
Issue #195 
There you go, lube nerfed for anyone other than traitors, maybe now nuke ops/traitors will buy the damn shoes sometimes, eh?
Doesn't slip on lube/soap/clown pda/etc. Go nuts.


DavidEdwards:
I dont see anything bad-balance wise. If anyone has a problem with this - you have a short time to say so before this will be on the master branch.

PR #221, opened 2016-02-03
New space suit sprites by Kmc2000

george99g:


Xantam-13:
Which space suits would this replace? Eva ones?

Time-Green:
Command Eva suits?

Alek2ander:
68747470733a2f2f692e6779617a6f2e636f6d2f38623531616436346337333631653764616537643632386231316364346265302e706e67

george99g:
> Which space suits would this replace? Eva ones?

It replaces the standard space suits from EVA, yeah.

Xantam-13:
It looks cyborgy... I'm not sure. Perhaps. I'm on the fence

Super3222:
It'd be pretty cool if these space suits could be in the commands EVA storage so we can keep the old ones... than we can have the best of both worlds.

Kmc2000:
@Alek2ander 

....

Why....

just why...

AsV9:
I'd like these as a new subtype of suits so both are there. Maybe some fancy features such as reagent view, security records checking overlay.

Kmc2000:
I like the command suit idea, that's probably the best way to add 'em as some people are unsure of them

ghost:
it looks like a space suit for cthulhu

just use other station's command suits

CMO's medical hardsuit is the white ship med hardsuit

CE has his hardsuit

HOS has his hardsuit

Cap has his  hardsuit

RD can get the bombsuit that's in /tg/

no offense but this looks like a strange crossbreed between an albino penis, a stormtrooper, and H.P Lovecraft

Kmc2000:
@AdamElTablawy You know those two suits on the right hand side of EVA that sit there, neglected and alone and never used?

That's what we're talking about

Also, THAT DESCRIPTION, LMAO

AsV9:
Making these a new subtype.

PR #233, opened 2016-02-04
Sleepers can be emagged

Super3222:
So this update will allow sleepers to be emaggable.
Idea from #167 and discussed with Adam, shortly.

When a sleeper is emagged, it will allow anyone to inject chemicals other than . Whether you are a medical doctor or not.

After it is emagged, if someone attempts to inject someone with one of the chemical choices they will be injected with SIXTY UNITS of every chemical choice available (depending on it's machinery tier). After this happens they will be popped out of the sleeper, and smoke will arise from the machine.

After this happens, the sleeper will enter a state where it is broken. This will prevent the traitor from doing the same thing over again in the same sleeper, and anyone else in the mean time. The sleeper will be unable to inject any more chemicals, read anyone's stat, etc. It will just provide a message informing the user a malfunction has occurred.

Is there a way to fix it? Yes there is! Close the sleeper, open up it's maintenance hatch with a screwdriver, and than use wire cutters on it to repair the damaged wires and you are good to go.

I've already ran many tests on this if you are wondering. Feel free to discuss.


X-TheDark:
Damn nice.

AsV9:
Might be OP but I'll merge this and we'll see what happens.

Issue #44, opened 2016-01-16
More character creation slots (from 3 to x)


george99g:
opened by Xantam · about a month ago
Nature:
More character creation slots (from 3 to x)
Thread linked:
http://forums.yogstation.net/index.php?threads/other-crazy_drakkon.7679/
Base content:
What is your idea type?:
Other

Give us a short description of your idea?:
Character Slots

Give us a full description of your idea?:
Pretty self explanatory. May sound kinda useless to some, but some may simply like more character loadouts than the three available.

AsV9:
This is part of what you get with BYOND on donating. I'm confused. Shouldn't we already have this for our donators?

Xantam-13:
I think the idea was that donors want maybe 5 slots?

X-TheDark:
Done.

PR #249, opened 2016-02-06
Extra character slots for donors (total of 6)

X-TheDark:
Issue #44 
As per title.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1203,4 +1203,4 @@ var/global/list/special_roles = list( //keep synced with the defines BE_* in set
 			character.update_body()
 			character.update_hair()
 			
-#define DONOR_CHARACTER_SLOTS
+#undef DONOR_CHARACTER_SLOTS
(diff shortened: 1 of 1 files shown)


Xantam-13:
If this works, I'll love you long tine.

X-TheDark:
Oh god, Travis is already on 510, huh?

X-TheDark:
Travis issue caused by it using 510 version to compile, which doesn't, due to Regex changes.

code/modules/client/preferences.dm
```diff
@@ -1198,4 +1201,6 @@ var/global/list/special_roles = list( //keep synced with the defines BE_* in set
 			hardset_dna(character, null, null, null, null, chosen_species, features)
 		else
 			character.update_body()
-			character.update_hair()
\ No newline at end of file
+			character.update_hair()
+			
+#define DONOR_CHARACTER_SLOTS
```
TehFlaminTaco: This should be undef, Not #define

PR #92, opened 2016-01-20
Dynamic User Interface

Cruix:
Requires testing
Response to issue #43

Adds:
-Support for custom Cyborg and AI UIs
-Support for donator-only UIs
-The ability to change your UI in-game through the "Select UI Style" verb in preferences
-Two new donator-only UIs, one for Cyborgs (Beepsky) and one for AIs (Matrix_OS)
-Three new saved preferences for UI style (one for each of AI, Cyborg, and Carbon)

Removes:
-the old method of selecting UIs in the pre-game lobby
-the old saved preference for UI style

Issues:
-UI always redraws when you change your carbon UI, even if you are currently using a different UI.
-When the UI redraws, there is a short period in which screen effects like blindness and flash are removed, and certain UI icons show incorrect values (such as the health doll)

Issue #43, opened 2016-01-16
New HUD's as optional


george99g:
opened by Xantam · about a month ago
Nature:
New HUD's and possibly GUI to select one
Thread linked:
http://forums.yogstation.net/index.php?threads/cosmetic-crazy_drakkon.7660/
Base content:
What is your idea type?:
Cosmetic

Give us a short description of your idea?:
Additional HUD styles.

Give us a full description of your idea?:
Other than a little spriting, I don't see the harm in additional HUD styles, and they're pretty cool. :P I'd recommend a Pipboy-esque and Syndicate style one? You say.

DavidEdwards:
New huds have been added.

PR #258, opened 2016-02-08
Dynamic ui

Cruix:
Requires testing
 Response to issue #43
 #92 Redone in a new pull request with fewer commits

Adds:
 -Support for custom Cyborg and AI UIs
 -Support for donator-only UIs
 -The ability to change your UI in-game through the "Select UI Style" verb in preferences
 -Two new donator-only UIs, one for Cyborgs (Beepsky) and one for AIs (Matrix_OS)
 -Three new saved preferences for UI style (one for each of AI, Cyborg, and Carbon)

Removes:
 -the old method of selecting UIs in the pre-game lobby
 -the old saved preference for UI style

Issues:
 -UI always redraws when you change your carbon UI, even if you are currently using a different UI.
 -When the UI redraws, there is a short period in which screen effects like blindness and flash are removed, and certain UI icons show incorrect values (such as the health doll)


AsV9:
> Issues:
> -When the UI redraws, there is a short period in which screen effects like blindness and flash are removed, and certain UI icons show incorrect values (such as the health doll)

Is this a major issue? Can this be exploited in a way?

Cruix:
If you are blind or flashed, you could spam-change your UI once per second to be able to see, but it would be difficult to play like that. I'm working on solutions, and while I'm at it I might be able to crunch all these tiny commits into one.

DavidEdwards:
Perhaps an overlay can be added, that covers everything in the interim time between UI updating.

AsV9:
Or could changing UI mode be disabled while the player is stunned maybe.

Cruix:
Right now I'm looking at only redrawing the parts of the UI that are redrawn, rather than the whole thing.

Cruix:
Closing this one more time. I will learn how to properly rebase and squash commits someday, but that day is not today.

PR #262, opened 2016-02-09
Adds the Chameleon Military Belt

### LINKED ISSUE #254: Chameleon Belt
Adds a chameleon military belt that you can use in-hand to change its appearance to any belt, which basically gives you 6 extra pockets that can be disguised. 

This is important because military belt is a great item but it's an immediate permabrig so nobody buys it ever. 

Disguise options- 

Toolbelt
Medical Belt
Security Belt
Fannypack
Janibelt
Military Belt for those badass traitors

Cruix:
Fixes #254 
Adds the Chameleon Military Belt as a traitor uplink item. It works exactly like the chameleon jumpsuit - it starts with the patterns for Toolbelt, Medical Belt, Security Belt, Fannypack(leather), Janibelt, and Military Belt, and patterns can be added by attacking the belt with a new belt. It flashes rainbow colors when hit by an EMP, but unfortunately the wearing sprite does not update, only the in-inventory sprite.

Costs 3-TC, the same as the old military belt, but has been moved from tools to stealth tools.

Removes the old military belt from the traitor uplink. Nuke ops still spawn with a normal one like they always did.

Fixes a bug in a chameleon jumpsuit message where the span class was escaped.

Fixes the description for no-slip shoes that should have been updated in #213.


X-TheDark:
Yup, totally forgot to update the uplink text. To make a note, it's not just the nuke shoes that not slip on anything, No-Slip Brown Shoes also do this.

PR #263, opened 2016-02-09
Just ported TG Lighting

TehFlaminTaco:
I'm sick of this engine. This works, I've tested it, As an added bonus,
even has my dim shit. Fuck yeah, Let me sleep.


DavidEdwards:
It is on master. Just needs to be moved to the server now.

Issue #82, opened 2016-01-19
Mod Jetpacks while they are on.


Super3222:
Make it where... if a player toggles their jetpack while it's on their back they can pass over tables.

shrimptk:
All jetpacks currently in ss13 are designed to only work in a weightless environment as a sort of 'guided thrust' system versus an actual jetpack. If this were to be implemented, it would have to only work if the requirements for the jetpack to function were met as well.

Silvaplex256:
That doesn't mean jet packs shouldn't do this with tables when the gravity gen breaks.

BadminBen:
Meh against this honestly. Pretend the ceiling isnt high enough to physically pass by while gilding,

Groudonmaster:
> Can climb on top of and run on top of tables
> Can't glide horizontally past them?
> Ben at least make sense.

PR #266, opened 2016-02-09
Flying over tables with jetpack

Super3222:
This will allow people who are using jetpacks (while they are on) to fly over tables whenever gravity is down in the area.

It works for everything defined as a jetpack, the captains own gold one, the carbon dixode one in EVA, the suit-inbuilt jetpack, etc.

Req. #82

### DIFF
--- a/code/game/objects/items/weapons/tanks/jetpack.dm
+++ b/code/game/objects/items/weapons/tanks/jetpack.dm
@@ -9,6 +9,7 @@
 	var/datum/effect/effect/system/ion_trail_follow/ion_trail
 	var/on = 0.0
 	var/stabilization_on = 0
+	var/flying = 0 // Needed to control the PASSTABLE flag
 	var/volume_rate = 500              //Needed for borg jetpack transfer
 
 /obj/item/weapon/tank/jetpack/New()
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/tanks/jetpack.dm
```diff
@@ -33,24 +34,36 @@
...
 	//	item_state = initial(item_state)
 		ion_trail.stop()
+		if(!has_gravity(C) && flying == 0 && C.pass_flags & PASSTABLE) // so it removes PASSTABLE which they get when its on. Prevents loopholes.
+			C.pass_flags -= PASSTABLE
 	usr << "<span class='notice'>You toggle the jetpack [on? "on":"off"].</span>"
 	return
 
-
 /obj/item/weapon/tank/jetpack/proc/allow_thrust(num, mob/living/user)
+	var/mob/living/carbon/human/C = usr
```
DavidEdwards: Does this need to be a human? What if for whatever reason, a different living creature manages to activate the jetpack. `has_gravity(C)` doesn't require a human type. If you can keep to using a living type, it would be best. If not, some sanity checks need to be made.

DavidEdwards:
Please also make sure that you add a changelog .yml file indicating your changes. There is an example here: /html/changelogs/example.yml.

code/game/objects/items/weapons/tanks/jetpack.dm
Super3222: @DavidEdwards Alright, sounds good. Though I never imagined mobs could wear or even activate jetpacks since they can't hold things. I was thinking that subspecies like adamnatine golems would come down under humans as well.

PR #272, opened 2016-02-11
Adds organ decay and makes heart transplants work

Cruix:
*Organs now have a decay timer that counts down when the organ's owner is dead or the organ is not in someone's body. This can be halted by putting the organ somewhere cold.
*Only hearts currently decay by default. They will not decay if the temperature is 0 degrees Celsius or colder. The decay time is 10 minutes, the same as the old defib time limit.
*Refrigerator lockers now cool down their contents just like freezer crates.
*Defibs now work off the new decay system instead of time of death, making heart transplants possible.

Inspired by this thread: http://forums.yogstation.net/index.php?threads/heart-transplant.8709/

### DIFF
--- a/code/game/objects/items/devices/scanners.dm
+++ b/code/game/objects/items/devices/scanners.dm
@@ -121,6 +121,17 @@ MASS SPECTROMETER
 	if(M.status_flags & FAKEDEATH)
 		mob_status = "<span class='alert'>Deceased</span>"
 		oxy_loss = max(rand(1, 40), oxy_loss, (300 - (tox_loss + fire_loss + brute_loss))) // Random oxygen loss
+		user << "<span class='danger'>Subject's heart tissue has decayed beyond the point of no return.</span>"//perhaps make this random too?
+	else
+		var/obj/item/organ/internal/heart/heart = M.getorgan(/obj/item/organ/internal/heart)
+		if(!heart)
+			user << "<span class='danger'>Subject does not have a heart.</span>"
+		else
+			if(heart.decay_time && (heart.decay != heart.decay_time))
+				if(heart.decay == -1)
+					user << "<span class='danger'>Subject's heart tissue has decayed beyond the point of no return.</span>"
+				else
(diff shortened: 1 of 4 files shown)


Time-Green:
This is amazing, now there is no limit for who you can revive!
Ok there is, since the cool new healing surgery got removed.

Maybe make reserves? But block them off to prevent people from eating them?

DavidEdwards:
Brilliant! I love seeing inventive PRs. 

Does this mean that people dead in space (read: frozen) can be brought back to life at any time?

Someone could take their suit off in space if they think they will die. That said, this is not necessarily a bad thing.

Cruix:
Storing dead bodies, or just hearts, in space will preserve the organ. I'm not sure if sitting dead in space causes tissue damage - if it does, that would prevent defibing. But if it does not, yes, dead bodies in space can be revived at any time.

Time-Green:
Can the healing surgery be re-added?

It would work amazingly well with this

Cruix:
I agree - it would be nice to be able to do surgery on almost any dead body to save it. It would provide a relatively costly (in that you would have to have the stored organs for the transplants and the consumable supplies for the healing surgeries) alternative for when cloning is backed up, or be a faster alternative to cloning as a reward for having a skilled doctor and a geneticist willing to give some humanized monkeys for hearts. Surgery is a fun but little-used part of doctoring, and I'd like to see it used more.

On a related note, if you look into the thread I linked to in the original comment, it talks about being able to grow artificial organs through a partnership between RnD and botany. By increasing the potency of the organ plants, botany could even grow organs with longer decay times than normal. I love this sort of optional but fruitful interdepartmental interaction, and would like to add it in the future.

Cruix:
I just added another commit that makes organ that are implanted work off of body temperature instead of air temperature. This means that space suits can now be used to insulate organs, like David says:

> Someone could take their suit off in space if they think they will die.

Organs in backpacks/pockets/etc. still work off ambient air temperature.

DavidEdwards:
I definitely feel more comfortable with that. Great work :)

PR #281, opened 2016-02-12
Bluespace Fanny Packs

TehFlaminTaco:
Adds bluespace Fanny Packs, which are like fannypacks, but bluespace.

Cannot be used to BoH Bomb.


DavidEdwards:
Looks good!

PR #286, opened 2016-02-13
Updated PDAs 

### LINKED ISSUE #273: Atmos PDA: provide atmos alert monitoring
Just like power monitoring in engineering PDA cartridge, so atmos wouldn't have to run back and forth just to see alerts.

Cruix:
Fixes #273 
*fixed some cases where the botanist's PDA would not act like a plant analyzer when it should. Removed some copy-pasted code.
*PDA cartridges can now receive atmos, fire, and power alerts. Atmos techs receive atmos and fire, Engineers receive power, and the CE and Captain receive all 3. Alert notifications can be toggled on, causing the PDA to beep and state alerts when they are received.
*Atmos tech, CE, and Captain cartridges can view all atmos monitoring consoles through their PDA, similar to how engineers can view power monitoring consoles.
*Changed scanmode magic numbers into #defines.
*Engineer PDA cartridges have a new scanmode that lets them measure power in wires like a multitool.
*Added some new menu categories, moved some options out of "Utilities".
*Captain's Value-PAK cartridge now has 100% more value.

I have tested this quite thoroughly, but I messed with so much bad PDA code that I would prefer lots of additional testing before it is pushed to live.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -14,6 +14,11 @@
 
 
 // ===
+
+var/list/power_alert_listeners = list()
+var/list/atmos_alert_listeners = list()
+var/list/fire_alert_listeners = list()
+
 /area
 	var/global/global_uid = 0
 	var/uid
@@ -72,10 +77,17 @@
 					D.cancelAlarm("Power", src, source)
 				else
 					D.triggerAlarm("Power", src, cameras, source)
+			for(var/listener in power_alert_listeners)
+				if(state == 1)
+					listener:cancelAlarm("Power", src, source)
+				else
+					listener:triggerAlarm("Power", src, cameras, source)
+
 	return
 
 /area/proc/atmosalert(danger_level, obj/source)
 	if(danger_level != atmosalm)
+
 		if (danger_level==2)
 			var/list/cameras = list()
 			for(var/area/RA in related)
@@ -88,6 +100,8 @@
 				a.triggerAlarm("Atmosphere", src, cameras, source)
 			for(var/mob/living/simple_animal/drone/D in mob_list)
 				D.triggerAlarm("Atmosphere", src, cameras, source)
+			for(var/listener in atmos_alert_listeners)
+				listener:triggerAlarm("Atmosphere", src, cameras, source)
 
 		else if (src.atmosalm == 2)
 			for(var/mob/living/silicon/aiPlayer in player_list)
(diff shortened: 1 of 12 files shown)


code/game/area/areas.dm
```diff
@@ -72,10 +77,17 @@
 					D.cancelAlarm("Power", src, source)
 				else
 					D.triggerAlarm("Power", src, cameras, source)
+			for(var/listener in power_alert_listeners)
+				if(state == 1)
+					listener:cancelAlarm("Power", src, source)
+				else
+					listener:triggerAlarm("Power", src, cameras, source)
```
X-TheDark: Please don't use this style. WHen I meant typeless loops, all I meant is
for(var/x in something)
var/typecast/here/please/variable = x

Alek2ander:
AI antags are unhappy

DavidEdwards:
What do you mean Alek?

DavidEdwards:
Please also make sure that you add a changelog .yml file indicating your changes. There is an example here: /html/changelogs/example.yml. (unless your recent PR contained that already)

Alek2ander:
> What do you mean Alek?

Right now antag AIs that do funny stuff (syphon rooms, vent plasma, etc) can get away with it because only supernerds check alerts on the console. But if every engineer has a PDA screaming at them, AI has to change the air alarm alert settings (which is clunky to do) if it wants to maintain any sort of stealth.

DavidEdwards:
I see. That is a good point. Worth consideration.

code/game/area/areas.dm
```diff
@@ -72,10 +77,17 @@
 					D.cancelAlarm("Power", src, source)
 				else
 					D.triggerAlarm("Power", src, cameras, source)
+			for(var/listener in power_alert_listeners)
+				if(state == 1)
+					listener:cancelAlarm("Power", src, source)
```
X-TheDark, line 82: Untyped variables.
SHUN! SHUUUUUUUUUUUNNNNNNNNNNNNNNNNNN!

Cruix, line 82: HUUUUUUUUSH this is one area where it makes sense to use the : operator, right now the alert_listener lists only contain cartridges, but they could very well contain everything in all 3 of those other horribly unmodular for-loops.

Cruix:
> Right now antag AIs that do funny stuff (syphon rooms, vent plasma, etc) can get away with it because only supernerds check alerts on the console

I don't think we should leave the game in such a state that antags think they can get away with things because most players are too lazy to do their job. What if traitors were willing to combo-cafe their target in the central hallway with plenty of witnesses because they know that sec is too incompetent to listen to their radio? But, if everyone disagrees with me, it will be easy enough to just remove the flags from everyone's PDAs.

I considered having alerts go through the PDA server, so the AI could turn that off, but decided against that because turning the PDA server off requires that the AI's camera go all the way to tcomms, which itself is very clunky.

Cruix:
I accidentally closed it...

DavidEdwards:
I think that we should go ahead with this, and see how it goes. We can always disable parts of it if it is too much.

Issue #255, opened 2016-02-08
Cable cuffs + wirecutter = cable coil ?


00aquir:
I suggest that we make it possible to make coils out of cable cuffs with a wirecutter. There is no other way to recycle it.

ghost:
Idea: allow cable cuffed people to cut their cablecuffs in ~10 seconds if they click on a wire cutter.
I think this your idea is good, aquir.

Super3222:
I think that if you are holding the cable cuffs in your hand, you should be able to untie it through activation.

Now if you use wire cutters on it while it's in your hand, I think that should just produce cable pieces.

ShadowDeath6:
Secret robust tipperino- You can attack someone who is cabecuffed or ziptied to instantly free them. 
Zipties turn into the cut form, while cabecuffs just vanish.

Super3222:
@00aquir How do you feel about people untying cable restraints by simply pressing Z (or double clicking it)?

00aquir:
I'm ok with using them in hand to make coils.

DavidEdwards:
Seems good in principle. I like Groxics idea too. If there is a cable cutter laying around, and you click on it when you are cuffed. You should be able to fiddle around until you get it in the right position to cut it.

Longer than 10 seconds though.

PR #290, opened 2016-02-13
You can now recycle cable restraints

Super3222:
#255.

Cable restraints can be recycled back into coils by untying them in your hand.

Furthermore, the other request will probably be in another PR. It was where if you're cuffed and click on a wirecutter, you can free yourself under a little over 10 seconds.

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -112,6 +112,13 @@
 			user << "<span class='warning'>You need one rod to make a wired rod!</span>"
 			return
 
+/obj/item/weapon/restraints/handcuffs/cable/attack_self(mob/user)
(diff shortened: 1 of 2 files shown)


DavidEdwards:
Looks good. Have you considered the implications of getting 15 coils back from a single restraint?

DavidEdwards:
Please also make sure that you add a changelog .yml file indicating your changes. There is an example here: /html/changelogs/example.yml.

AsV9:
@DavidEdwards Don't cable coils require 15 lengths of cable to make?

Super3222:
Cable restraints require 15 lengths of cable to make. I made it so the new cable coil amount that you get equals 15.

> Please also make sure that you add a changelog .yml file indicating your changes. There is an example here: /html/changelogs/example.yml.

Thank you, I'll start doing that for the rest of my PRs. The example was helpful.

Issue #23, opened 2016-01-16
NTSL cancels AI Tracking


george99g:
opened by Nicholas Weber · 3 months ago
Adding NTSL Scripts bugs out AI tracking, and will not allow you to track people at all.
Source:http://forums.yogstation.net/index.php?threads/ntsl-scripts-ai-camera-tracking.6663/
Additional Sources: I've done it myself to test its theory, Adding anything even a job indicator usually bugs it's tracking and will not allow you to track at all.

Groudonmaster:
That's because it tracks by searching for the name defined but if you add a job title to the name the script can't locate it because instead of searching for "John Sweeps" it searches for "[Sci]John Sweeps"

oisin100:
Fixed in PR

PR #293, opened 2016-02-13
Fixes NTSL Ai tracking.

oisin100:
AI no longer uses the players name to track them.

Instead it uses a new NTSL variable "$uuid" whats exactly the same as the name but makes sure the AI can still track the player even if they change their name with a NTSL script.


Groudonmaster:
Bloody finally. I didn't want to create a whole new id just for AI tracking but I guess you had too

oisin100:
Should probably ref this... #23

Issue #302, opened 2016-02-14
Botany : Drying racks


ToGWtF:
Drying racks turn invisible when screwdrivered, have to unwrench to make them visibile by Alt+Clicking the tile its on. Can be used to make invisible walls

oisin100:
Fixed.

PR #304, opened 2016-02-14
Fixes invisible drying racks.

oisin100:
All in the title. Drying racks have no maintenance icon so i just set it to "drying_rack"

 #302


DavidEdwards:
Great, needs a changelog .yml though

oisin100:
Done.

Issue #297, opened 2016-02-14
Shadowling Thrall Darksight


ghost:
As a thrall, nothing happened when I used darksight one round. I don't know if this is a consistent bug, though.

Cruix:
Seems the same as #39 
Were you wearing any kind of glasses? That prevents darksight from working.

PR #310, opened 2016-02-15
Adds glasses warning to Thrall Darksight ability

### LINKED ISSUE #39: Shadowlings/Thralls
opened by Ben · 2 months ago
What is the nature of the issue: It seems that thralls are having trouble activating Dark vision. Got about 3 ahelps from it. They are stating they got turned into a thrall and when attempting to use the vision it wont activate
What steps must be taken to reproduce this error:
Any additional information? : Was tested and it works. Thing is it will not work if you have glasses on. I tested with sunglasses maybe other glasses will allow the ability to be used.
If you want to keep it this way i recommend a description be put that glasses block your sight or something to that effect

Cruix:
Adds a warning to the thrall darksight ability, informing the user that glasses will block the ability. Will hopefully cut down on thrall ahelps.

If necessary I could instead make darksight work even with glasses on, but I would have to rewrite glasses code a bit and might run into problems.

Fixes #39 and #297.

### DIFF
--- a/code/game/gamemodes/shadowling/shadowling_abilities.dm
+++ b/code/game/gamemodes/shadowling/shadowling_abilities.dm
@@ -233,7 +233,7 @@
 		if(istype(H.glasses, /obj/item/clothing/glasses/night/shadowling))
 			var/obj/item/clothing/glasses/night/shadowling/eyes = H.glasses
 			if (eyes.darkness_view == 0)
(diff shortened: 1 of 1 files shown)


FluffySurvivor:
I don't see why we should legitimize a bug. Darsksight should work with glasses.

ghost:
It's not a bug though.

It's intended behavior.

DavidEdwards:
Ah! This one needs a yml changelog!

Issue #271, opened 2016-02-10
Buff Syndieborgs


ShadowDeath6:
Even with a cost of 80 TC, syndieborgs aren't exactly good. 

LMG is literally a slower firing revolver with a higher ammo capacity. 
Mini ebow got removed
Emag got removed

I'd like for the LMG to get a firing speed boost or let it fire in bursts like the ACTUAL syndie LMG. 
Give it a mini ebow for a decent sidearm while the LMG regains ammo
Give it an emag so it can hack stuff like other borgs. 

If none of these sound reasonable, lower TC cost to 70 or 60 to encourage it actually being used since nobody ever buys it. 

Tl;dr: Shadow gets buttmad at tg for making syndieborgs suck

DavidEdwards:
Everyone, what are the thoughts on this?

Time-Green:
Your buying an expensive murder machine. Not a mad kitten.

They will still fail from ion and flashes.

And perhaps give it a crowbar to open the other borgs and hack them

Still, just walk into AI core and emag him should be enough

ShadowDeath6:
> perhaps give it a crowbar

It has a crowbar still, from when it had an emag, so all it would need is the actual emag to be able to hack other borgs.

Time-Green:
Sorry, never been a syndiborg

ghost:
Giving it an emag'd be nice.

It could also spawn with an inbuilt VTEC module, and perhaps come with a 'syndiborg repair kit'-AKA cyborg repair module?

oisin100:
It doesn't have a VTEC module? Why the fuck doesn't it have a VTEC Module?

Giving it a emag could work well i wonder why they removed it...
Ebow would be an extremely useful sidearm whilst your LMG reloads.
As for it being flashed maybe it could get some sort of "digital light filter" what can be toggled on and off but uses a large amount of energy.

X-TheDark:
I'd at least buff the reload rate of the LMG (currently 1 bullet per 5 seconds) also what's this cyborg repair module?

Also, you know what'd be !FUN!, letting Syndiborgs reload their Grenade Launcher (at the cost of energy, of course, similar to mech weapons).

ghost:
no

flashes and EMP's are their weakness PERIOD

NO CHANGES TO THIS IT IS BALANCED

shrimptk:
Lets finalize gear changes

For a whoppin' 80 TC, you get a syndieborg with:

> VTEC Module 
[...]

PR #314, opened 2016-02-15
Syndiborg buff

X-TheDark:
Issue #271 

Syndiborgs now come with plethora of buffs:
1. LMG recharges every 2 ticks (instead of 5), supports firing modes of 1/3/5 rounds. Examining the Cyborg LMG now tells you how much ammo it has in it. Maximum capacity is 30 (not changed, just readily apparent now).
2. Grenade Launcher replicates 1 shot every 6 seconds when activated and standing still (500 energy per replicated shell);
3. Emag added (no energy usage);
4. Crowbar changed to red (no energy usage still);
5. Energy Sword cost down to 150 (from 500);
6. VTEC module comes added;

Flash is unchanged, still works the same way it did. 
As a sidenote, after this PR, energy weapons will support burst_size properly.

I expect this to make Syndicate Cyborgs completely broken.

**ANY ADDITIONAL COMMENTS/THINGS TO ADD WELCOME**

### DIFF
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -114,7 +114,7 @@
 	return
 
 /obj/item/weapon/melee/energy/sword/cyborg
-	var/hitcost = 500
+	var/hitcost = 150
 
 /obj/item/weapon/melee/energy/sword/cyborg/attack(mob/M, var/mob/living/silicon/robot/R)
 	if(R.cell)
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -1149,11 +1149,12 @@
 
 /mob/living/silicon/robot/syndicate/New(loc)
 	..()
-	cell.maxcharge = 25000
-	cell.charge = 25000
+	cell = new /obj/item/weapon/stock_parts/cell/hyper(src)
 	radio = new /obj/item/device/radio/borg/syndicate(src)
 	module = new /obj/item/weapon/robot_module/syndicate(src)
 	laws = new /datum/ai_laws/syndicate_override()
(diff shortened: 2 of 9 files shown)


ShadowDeath6:
Tfw no ebow

Looks good, syndieborgs SHOULD be broken. They suck right now.

Alek2ander:
> syndieborgs SHOULD be broken

They are already broken.
Additional teammate who is unstunnable by normal weapons, blocks weapon fire, suffers no slowdowns, is always fire and spaceproof, has all access, and can solo-destroy the AI. Apparently that's not OP enough.

X-TheDark:
What do you mean blocks weapon fire? Unless you mean acting as a bullet magnet, other than that, it doesn't reflect any bullets.

DavidEdwards:
> has all access, and can solo-destroy the AI

That is certainly dangerous. Why is it never taken?

DavidEdwards:
> has all access, and can solo-destroy the AI

That certainly is dangerous. Why is it never taken?

ShadowDeath6:
> Why is it never taken?

Because it sucks against fighting crew. An 80 TC AI kill isn't worth it when you can buy 2 c4 and an eshield for the same effect.

ghost:
It's terrible V crew, right now.

Not to mention, the moment ops hit the station and they're declared, AI just turns off environmental control, GG.

The AI core we have is in the center of the station, not on a sat like every other core, meaning gee fucking gee.

Not to mention that borgs are only ever bought during WAR!, where the crew knows they're coming and all have flashes.

X-TheDark:
Either way, the borg should have options now, cutting bolts if AI dicks him over (doesn't use energy) and crowbaring while defending itself with infinite LMG.

X-TheDark:
Reverted the infinite ammo LMG, back to regular energy consumption, just faster recharge (2 ticks instead of 5).

PR #319, opened 2016-02-16
Nuke Ops now actually have 5 minutes (instead of 2) to decide on War

X-TheDark:
Title. Changed the time to make War from 5 minutes to 8, because round starts at 12:03 which caused ops to have only 2 minutes to make an announcement.


FluffySurvivor:
:+1:

Issue #295, opened 2016-02-13
Drone hats on bodies


ToGWtF:
If a person picks up a drone, puts it on his head and dies,  the drone cannot resist off his head and is stuck there

PR #320, opened 2016-02-16
Drones no longer get stuck on heads after holder dies

X-TheDark:
Issue #295

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
@@ -41,10 +41,16 @@
 
 	if(istype(loc, /mob/living))
 		var/mob/living/L = loc
-		L.show_message("<span class='notice'>[drone] is trying to escape!</span>")
-		if(!do_after(L, 50, target = L) || loc != L)
-			return
-		L.unEquip(src)
+		if(L.stat == DEAD)
+			src.forceMove(get_turf(src))
+			if(L:head && L:head == src) //Guess whos time it is to be shunned?
+				L:head = null
+				L.update_inv_head()
+		else
+			L.show_message("<span class='notice'>[drone] is trying to escape!</span>")
+			if(!do_after(L, 50, target = L) || loc != L)
+				return
+			L.unEquip(src)
 
 	contents -= drone
 	drone.loc = get_turf(src)
--- a/code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
@@ -43,7 +43,7 @@
 		var/mob/living/L = loc
 		if(L.stat == DEAD)
 			src.forceMove(get_turf(src))
-			if(istype(L, /mob/living/carbon)
+			if(istype(L, /mob/living/carbon))
 				var/mob/living/carbon/C = L
 				C.head = null
 				C.update_inv_head()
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
```diff
@@ -41,10 +41,16 @@
 
 	if(istype(loc, /mob/living))
 		var/mob/living/L = loc
-		L.show_message("<span class='notice'>[drone] is trying to escape!</span>")
-		if(!do_after(L, 50, target = L) || loc != L)
-			return
-		L.unEquip(src)
+		if(L.stat == DEAD)
+			src.forceMove(get_turf(src))
+			if(L:head && L:head == src) //Guess whos time it is to be shunned?
```
Cruix: Nooooooooooooo shuuuuuuuun.
This is not a loop looping through a list of potentially many-typed objects and calling the same method on all of them, this is an if statement that you could easily and efficiently add a typecast to. This is not a place for the : operator. If a non-carbon picks up a drone, there will be a no-such-variable exception and the call will crash, making it impossible for drones to escape from any dead non-carbons. Just add an istype(L, /mob/living/carbon) to the if and typecast it.

X-TheDark: The if(L:head) part of the if prevent this from crashing, since if it doesn't exist, this will just false out without trying to access null (&& short-circuits in BYOND).

The reason I am doing this is because there are multiple things with hats (not all with the same subtype), although realistically, only humans can pick up drones and stuff.

Cruix: I was under the impression that the : operator caused a runtime error when it attempts to access a var that did not exist. If it just returns false, that's fine then.

DavidEdwards:
You are correct Cruix: 

> This is the runtime search operator. It is used to access a property of a var that is not explicitly prototyped. If the variable doesn't have the specified variable, a run-time error occurs.

code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
DavidEdwards: You are correct Cruix:

> This is the runtime search operator. It is used to access a property of a var that is not explicitly prototyped. If the variable doesn't have the specified variable, a run-time error occurs.

code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
```diff
@@ -41,10 +41,17 @@
...
 		var/mob/living/L = loc
-		L.show_message("<span class='notice'>[drone] is trying to escape!</span>")
-		if(!do_after(L, 50, target = L) || loc != L)
-			return
-		L.unEquip(src)
+		if(L.stat == DEAD)
+			src.forceMove(get_turf(src))
+			if(istype(L, /mob/living/carbon))
+				var/mob/living/carbon/C = L
+				C.head = null
```
Cruix: You should re-add the if that checks that the drone is the thing on the carbon's head, otherwise a drone leaving a dead body will delete that body's hat, even if it shouldn't.

code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
X-TheDark: Oh, cool, didn't know that, was under impression that it would false if no variable.

code/modules/mob/living/simple_animal/friendly/drone/drones_as_items.dm
X-TheDark: Will do.

DavidEdwards:
Looks great! I'm really happy to see you both checking each others PRs. Keeps the quality high!

PR #329, opened 2016-02-18
Reduces PDA alert spam, especially in the case of fire alarms.

Cruix:
Rooms with fires had a tendency to trigger fire alerts rapidly, which would spam ghosts and anyone near a PDA set to receive fire alerts. This would also fill the "view alerts" screen with the same area multiple times. This fixes the issue by making PDAs not accept any alerts if the alert's area is already stored in the PDA's alert buffer.

### DIFF
--- a/code/game/objects/items/devices/PDA/cart.dm
+++ b/code/game/objects/items/devices/PDA/cart.dm
@@ -285,19 +285,19 @@
 	var/msg
 	switch(class)
 		if("Atmosphere")
-			if(!(alert_flags & PDA_ATMOS_ALERT))
+			if(!(alert_flags & PDA_ATMOS_ALERT) || (A in atmos_alerts))
 				return
 			atmos_alerts += A
 			if(alert_toggles & PDA_ATMOS_ALERT)
 				msg = "Alert: Atmos alarm in \the [A]"
 		if("Power")
-			if(!(alert_flags & PDA_POWER_ALERT))
+			if(!(alert_flags & PDA_POWER_ALERT) || (A in power_alerts))
 				return
 			power_alerts += A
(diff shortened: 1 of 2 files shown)


code/game/objects/items/devices/PDA/cart.dm
```diff
@@ -285,19 +285,19 @@
...
 		if("Power")
-			if(!(alert_flags & PDA_POWER_ALERT))
+			if(!(alert_flags & PDA_POWER_ALERT) || (A in power_alerts))
 				return
 			power_alerts += A
 			if(alert_toggles & PDA_POWER_ALERT)
 				msg = "Alert: Power alarm in \the [A]"
 		if("Fire")
-			if(!(alert_flags & PDA_FIRE_ALERT))
+			if(!(alert_flags & PDA_FIRE_ALERT) || (A in fire_alerts))
```
X-TheDark, line 300: Do you think it would be a better option to replace atmos/power/fire_alerts to an assoc list, so instead of
`(A in alert_type)`
You could do something like
`!isnull(alert_type[A.name])`
to prevent iterating over the entire list all the time.

This is a moot point, in my opinion, since the lists are not nearly big enough to warrant this approach, but I would like to hear your thoughts on this.

Cruix, line 300: That would be more efficient, since I think DM uses a binary search for associative string lists. However, if an alarm were triggered and the trolling CE used his blueprints to change the name of the room, that alarm would remain in the "view alerts" tab forever, until he changed the name back and the alert got cancelled. It would also prevent multiple alerts from different rooms with the same name, which could be a good or bad thing.

Issue #334, opened 2016-02-18
Dead Blobs being able to respawn?


Rocks4Free:
Today I was playing as a Blob, and I got bombed and died. My game disconnected for whatever reason, and when I reconnected, I was able to join the game again, even though I had already bursted and died as blob. Not sure if this is intentional or what, but just in case it isn't I'm reporting it here.

Cruix:
Repeat of #36 and possibly #32. Can we fix this by making blobs be ghosts when they die?

Groudonmaster:
Again. The issue is that when the blob core is destroyed it deletes the overmind mob. Find where that proc is and change it to gib the overmind instead. Random gibs out of nowhere but it should cause a ghost.

Time-Green:
It never does. I have been blob 4 times, including sub conscious.
I died 2 times and both of them I got kicked and respawned

DavidEdwards:
This issue should be fixed by the recent blob changes. However, if issues persist. Please make a new issue (you can reference this one).

PR #337, opened 2016-02-18
Blob Improvements

X-TheDark:
Issue #334
EDIT:

**Blob together!**
- If your core dies but there is still another one around (Split Consciousness), you do not die. You can help out the other Blob by building stuff. You earn resources from your Resource Nodes normally.
- Split/Relocate do not work if you do not have a core.
- When the last core dies, all active overminds will become ghosted. No more being kicked out of the server (**if necessary, tell me to separate this change from the rest of the stuff**)
- Small optimizations to explosion code. We already get only turfs from **trange**, no need to declare a type.

EDIT The Second : ~~Stealing~~ Porting code from /tg/ by @ChangelingRain 
tgstation/-tg-station#11495 - Blobs made explosion resistant, core immune to explosions.
tgstation/-tg-station#13516 - Blobs can block atmos, Strong Blobs and Core do so. Split Consciousness requires a Node under it, instead of finding the oldest one.
tgstation/-tg-station#15483 - Blobs can expand on lattices.

### DIFF
--- a/code/game/gamemodes/blob/overmind.dm
+++ b/code/game/gamemodes/blob/overmind.dm
@@ -34,11 +34,7 @@
 	ghostimage = image(src.icon,src,src.icon_state)
 	ghost_darkness_images |= ghostimage //so ghosts can see the blob cursor when they disable darkness
 	updateallghostimages()
-	..()
-
-/mob/camera/blob/Life()
-	if(!blob_core)
-		qdel(src)
+	blob_overmind_list += src
 	..()
 
 /mob/camera/blob/Destroy()
@@ -47,6 +43,7 @@
 		qdel(ghostimage)
 		ghostimage = null;
 		updateallghostimages()
+		blob_overmind_list -= src
 	//..()
 
 /mob/camera/blob/Login()
--- a/code/game/gamemodes/blob/blob.dm
+++ b/code/game/gamemodes/blob/blob.dm
@@ -4,6 +4,7 @@
 var/list/blobs = list()
 var/list/blob_cores = list()
 var/list/blob_nodes = list()
+var/list/blob_overmind_list = list()         //blob overmind list so we can keep them alive until all cores are dead.
 
 
 /datum/game_mode/blob
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -36,6 +36,7 @@
 /obj/effect/blob/core/Destroy()
 	blob_cores -= src
 	if(overmind)
+		overmind.ghostize()
 		overmind.blob_core = null
 	overmind = null
 	SSobj.processing.Remove(src)
--- a/code/game/gamemodes/blob/blobs/core.dm
+++ b/code/game/gamemodes/blob/blobs/core.dm
@@ -4,11 +4,13 @@
 	icon_state = "blank_blob"
 	health = 200
 	fire_resist = 2
+	atmos_block = 1
 	var/overmind_get_delay = 0 // we don't want to constantly try to find an overmind, do it every 30 seconds
 	var/resource_delay = 0
 	var/point_rate = 2
 	var/is_offspring = null
 
+	
 /obj/effect/blob/core/New(loc, var/h = 200, var/client/new_overmind = null, var/new_rate = 2, offspring)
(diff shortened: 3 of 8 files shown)


Cruix:
Will this cause any issues with the split consciousness ability? Can different overminds command each others spores and such? Perhaps overminds should only be ghostized if there are no blob cores left on the map.

X-TheDark:
That is a good point, I should probably test this with multiple overminds with the Split thing.

X-TheDark:
The problem is, though, that the Overmind is attached to the Blob Core, so if that get's Qdel'ed, You're done. Either ghostizie or get kicked and I'm pretty sure we don't keep Overminds in a separate array or anything.

Cruix:
Just add a new array in code/_globalvars/lists/mobs

ShadowDeath6:
Holy shit, having an overmind continue living if there's another core sounds amazing, might be kind of super powerful if there's another blob somewhere else, but :+1: 

Also, maybe add the verb for blob telepathy before bursting? Coordination makes for good blobs.

code/game/gamemodes/blob/overmind.dm
```diff
@@ -34,11 +34,7 @@
 	ghostimage = image(src.icon,src,src.icon_state)
 	ghost_darkness_images |= ghostimage //so ghosts can see the blob cursor when they disable darkness
 	updateallghostimages()
-	..()
-
-/mob/camera/blob/Life()
-	if(!blob_core)
-		qdel(src)
+	blob_overmind_list += src
```
Cruix, line 37: Won't this put the mind in the list over and over again, with every tick of Life()? Shouldn't you put this in New()?

EDIT: Ignore this, it is in New(), I just can't read the single-column compare view apparently.

B1gb3ast:
So bombs become useless vs blobs.

HUH

ghost:
that was already a thing on /tg/

blobs were bomb resistant there

code/game/gamemodes/blob/overmind.dm
X-TheDark, line 37: Yeah, I removed Life() proc as it no longer needs to exists, since Overminds don't die when you lose your core we don't check for core every tick.

X-TheDark:
Not useless. Basically, you cannot kill the blob/core via a bumrush done by 1 person. Also, explosion code is such that targets behind something resistant do not get protection, so if you have normal blob behind shield blob, shield blob will withstand (depends on explosion strength) but the normal blobs die.

The biggest gains you will notice will be against small explosions. Maxcap will still destroy blob (except core) very effectively.

Also, Split is such an expensive ability that I really thought it needed **something**. And not losing all your resources and nodes after death (the other guy doesn't get them, so all resource nodes basically become useless after your death) might just be that **something**.

ShadowDeath6:
Is this adding new blobchems, or do we keep old ones?

ghost:
/tg/ made bombs entirely useless against blobs

add that instead of a harmless nerf that does nothing

X-TheDark:
Unless I misunderstand something and /tg/ changed how ex_act is processed somewhere down the stack **all these numbers and effects were taken from the latest /tg/ blob code**, so it should be the same as it is there in regards to bombs and shit.

Also, for now, this is oldblob chems, I'll see if I can get around to adding new chems in a different PR, as this one has become bloated enough already.

FluffySurvivor:
A bunch of buffs with no compensation. The lattice buff is huge as you can destroy space emitters much easier.

ShadowDeath6:
Blobs literally never win anymore due to one shitler running in and maxcap bombing it. This is a much needed buff.

ghost:
Not to mention that not every buff needs a nerf to accompany.

/tg/ pulled it off fine, so can we.

ShadowDeath6:
Would this allow any exploits along the lines of a blob being able to respawn its core via the relocate core ability?
Edit: I didn't read the OP fully, oops

X-TheDark:
> - Split/Relocate do not work if you do not have a core.

Thanks, Adam, for pointing this one out by the way.

ShadowDeath6:
Shit, didn't notice that.

ghost:
MERGE

ShadowDeath6:
> MERGE

Issue #25, opened 2016-01-16
BuiltBy var


george99g:
opened by Alex West · 2 months ago
A var on machines/girders/walls/grilles to see who last did something in their construction, or maybe just a log of everyone who did any part in the construction of it. Would be very helpful, seeing as fingerprints fail to do that.

Super3222:
I'm onto this.

PR #346, opened 2016-02-20
Builtby Variable

Super3222:
This will add a "builtby" variable, which will allow admins to see who exactly was associated with creating objects, structures, machinery, etc. It will also differentiate between those loaded with the map and those created in game.

I'll do the same for craftable items/weapons after ghetto armoury gets merged.

How to get to it: View Variables > Type next to "Search:" builtby > It displays the users ckey and the users current in game name.
#25 requested this.


code/game/machinery/machinery.dm
```diff
@@ -124,6 +124,12 @@ Class Procs:
 	machines += src
 	SSmachine.processing += src
 	power_change()
+	var/mob/user = usr
+	if(ticker.current_state == GAME_STATE_PLAYING)
+		src.builtby = "This was created by [user.ckey]/[user.name]."
```
Cruix, line 129: You should check to make sure the user isn't null, just in case.

ghost:
OH THANK GOD

Alek2ander:
Why use a separate variable when you can put this info in fingerprints?

Super3222:
Eh, I might close this for two reasons.
1. Fingerprints basically do the same thing. We really don't need this.
2. Some things in the code may show up as null unexpectedly after construction, and there's no way around it.

PR #349, opened 2016-02-20
Warden secway Keys

oisin100:
Secway key now spawns inside wardens closet.

Changelog contains < strike > tags they better be compatible..
#349


AsV9:
> Changelog contains < strike > tags they better be compatible..

I don't believe it before I see it.

oisin100:
Im not sure either. But im assuming when it builds the changelog it will compile it as pure HTML so the < strike > tag should work..

PR #352, opened 2016-02-21
Odysseus Improvements, Part 2.1 : Rearming Syringes

X-TheDark:
Yup, this is still not dead yet.

Syringes can be rearmed now:
- 3 seconds per syringe;
- Can only be done in bulk (all or nothing);
- 250 energy per syringe : at 10, you are looking at 12.5% energy of the standard 20k cell mechs get.
- Must stand still for entire duration;

All of the above are purely for balance. I'd like rearm to be instant or have Syringe Gun not have ammo but generate syringe when firing.

### DIFF
--- a/code/game/mecha/equipment/tools/medical_tools.dm
+++ b/code/game/mecha/equipment/tools/medical_tools.dm
@@ -231,6 +231,7 @@
 
 ///////////////////////////////// Syringe Gun ///////////////////////////////////////////////////////////////
 
+#define SYRINGE_REP_TIME 30 //Time (in deciseconds) to rearm 1 syringe
 
 /obj/item/mecha_parts/mecha_equipment/syringe_gun
 	name = "exosuit syringe gun"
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
:+1: x :100: 
Why use so much energy though? Making 300 bullets costs less.

X-TheDark:
Balance. I would rather decrease the costs later, if needed. Also, keep making suggestions even in here. If you think this should cost less, tell me and I will do it if enough people think the same way.

Plus, you can pick the syringes back up. This is mostly added as **an option** (something Odysseus severly lacks in the syringe department).

Hell, if people want me to, I can just remove the ammunition on Syringe Gun entirely and make it shoot as long as you have energy (basically like laser guns and stuff).

ShadowDeath6:
Maybe just make it have an infinite amount as long as it has chems to fire.

X-TheDark:
Literally the reason for all this needless expense and time to rearm syringes is to prevent their spam, since, unlike lasers, one hit from this can be extremely OP, not mentioning the fact that if you use some stam-damaging chem, you can walk up to them and start punching them and they will not be able to do anything (since mech punches paralyze).

ShadowDeath6:
Non-combat mechs don't stun on punch, I'm pretty sure.

DavidEdwards:
Looks good! See how it runs in game when it is live. You can tune the numbers if it feels too sluggish.

PR #355, opened 2016-02-21
Rezadonut box for xenobiology.

B1gb3ast:
So lazy scientists won't have to go to medbay, woo!
Convenience!
Easy mode!
Screw cryo!

### DIFF
--- a/code/game/objects/items/weapons/storage/fancy.dm
+++ b/code/game/objects/items/weapons/storage/fancy.dm
@@ -51,6 +51,10 @@
 	can_hold = list(/obj/item/weapon/reagent_containers/food/snacks/donut)
 	spawn_type = /obj/item/weapon/reagent_containers/food/snacks/donut
 
+/obj/item/weapon/storage/fancy/donut_box/rezadone
+	name = "rezadonut box"
+	spawn_type = /obj/item/weapon/reagent_containers/food/snacks/donut/rezadone
+
 /*
  * Egg Box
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
:+1:

ghost:
these kill people

these legit kill people because there's too much rezadone in them and if they eat 3 donuts they are fucking dead

also

why are you removing department interdependance

if you're going this way just add cyro to each department, because god forbid people actually leave their areas to interact with people or depend on other departments for assistance

bad.

ShadowDeath6:
If you die to rezadone you suck

These are for lower damage problems, like being latched by a slime for a second or two. This will in no way replace cryo.

B1gb3ast:
Can't wait for some tresspassing clown to munch down on all the donuts.

Should I add a "mildly toxic" to item description?

There's only 15 units of rezadone in each donut, and from what I have heard it's 1 damage per unit.

ShadowDeath6:
Maybe have the box say something like "Note- Eating multiple Rezadonuts is highly discouraged by our experts and may result in disfiguration of facial tissue. "

Time-Green:
Just add it. Only for the idea of giving it to the clown

DavidEdwards:
I am against removing inter-department cooperation mostly. That's one reason we "broke" the grinder.

Unless there is an uproar demanding such a thing, I'll close this PR for now.

PR #369, opened 2016-02-23
Disables editing of sound related variables unless admin has +SOUND.

AsV9:
There are probably more abusable variables but this has got to do for now.


Wardog390:
i take it your making sure only +SOUND can counter this line of code?

AsV9:
Currently nobody. I could do that though.

AsV9:
Voila.

Wardog390:
seeing as we are sucking all the fun, you might as well spend the time looking up sounds that can be changed in VV and limiting those, getting +SOUND has to be worth it.

AsV9:
To be honest nobody should be touching things they're not supposed to touch. That shouldn't be too much to ask for.

Issue #247, opened 2016-02-06
Poster Abort Bug


ghost:
After aborting the placement of a poster, the poster gets rolled up again and dropped on the floor. When the poster is being placed again on any wall, none of the icons will show up, including the poster being placed and the poster on the wall, despite the poster being on the wall and not being able to place new posters on that wall.
![1](https://cloud.githubusercontent.com/assets/16929778/12864650/a30a8ae6-cc5f-11e5-8b6a-c315d0b61e4f.jpg)
![2](https://cloud.githubusercontent.com/assets/16929778/12864651/a6ec1706-cc5f-11e5-9a63-313ae35279f5.jpg)
In the images, the poster is being placed after being aborted and now icons appear.

ghost:
In the last sentence, "now" is supposed to be "no." I believe this bug occurred before Pull Request #209.

DavidEdwards:
Since you wrote the last thing to do with posters, perhaps it would be good for you to investigate this?

PR #383, opened 2016-02-24
Poster Fixes

Cruix:
Fixes issue #247
Fixed posters not being placeable after failing to place them once. Also made poster code use a do_after() instead of sleep().


Cruix:
If anyone is interested, it was because aborting a poster put the poster's structure's loc in the item, and when the item gets qdel'ed, the structure gets deleted because it is in the item, whereas when a poster was New()'d, the structure's loc was set to null.

Issue #294, opened 2016-02-13
Bandage duplication


ToGWtF:
This is only a graphical glitch, as you don't end up with two useable bandages

> Begin to apply bandage
> Switch hands
> You switched hands, so you drop bandage because you failed to put it on
> But you drop bandage from a hand that was empty, leaving the real bandage in your other hand
> Pick up bandage on ground, bandage switches hands
> The hand the bandage was in now has an invisible bandage
> Use bandage from either hand, it uses both up

Also, you can use a bandage from an empty hand if you have it in the opposite hand. Not sure if this behavior was intentional.

Cruix:
> Also, you can use a bandage from an empty hand if you have it in the opposite hand. Not sure if this behavior was intentional.

I am unable to reproduce this, can you give more details?

ToGWtF:
Uh. If you hold a bandage in one hand, duped or not, you can switch to your empty hand to apply it to yourself

PR #385, opened 2016-02-24
Fixes bandage visual duplication bug.

Cruix:
Fixes the first part of #294. I was unable to replicate the second part.

Now, if you switch hands or place a bandage somewhere while you are applying it, you do not drop it. You still drop it if the person moves or you are stunned. Added some "\improper"s to bandage names.


ToGWtF:
Yeah. I think the second part of the issue was me not realizing it left an invisible one in your hand at first

PR #389, opened 2016-02-25
Fixes job temp bans

oisin100:
Changed the SQL syntax to be like normal bans. This should fix the tempbans not lifting

If someone could actually test this for us in a production environment it would be nice...

PS: Dont make me go into detail on why this fixes it... Just look at coder chat...


oisin100:
Oh god. My git bugged out so hard uploading that code...

PR #390, opened 2016-02-25
Fixes job temp bans

oisin100:
Title says it all.

Pretty much the old code was "(bantype = 'JOB_TEMPBAN' AND (isnull(unbanned) OR expiration_time > Now()))"

Which meant that tempbans would not auto lift unless the job tempban is removed. Pretty much making job tempbans = job permabans


oisin100:
Also, its untested. If someone could test it in the dev environment for me it would be sweet.
The code only uses the "SELECT" syntax so its perfectly safe to test.

DavidEdwards:
I'll test this when I get home.

oisin100:
To test: Jobban yourself for one minute. It should lift automatically after a minute. If not then i fucked up.

PR #404, opened 2016-02-27
Fixes Preternis Nutritious Reagent Behavior

Cruix:
Fixed Preternis not metabolizing any nutritious reagents. They now metabolize them normally except that they receive 1/5th the nutrient from them, as intended. The message they receive when they eat food has been changed to appear once every two minutes if there is a nutritious reagent in their system, instead of on a 1% probability per tick there is a nutritious reagent in their system.


DavidEdwards:
Ah! Can you add a changelog?

PR #405, opened 2016-02-27
isnull is Required!

oisin100:
Bans not fixing. Coder chat assuming its something to do with unbanned != 1. In SQL NULL cannot be compared to other objects such as 1. This may be causing the connection from the server to the database to be killed. This is all in theory.

Theres a reason why isnull exists.

### DIFF
--- a/code/modules/admin/banjob.dm
+++ b/code/modules/admin/banjob.dm
@@ -40,7 +40,7 @@ var/jobban_keylist[0]		//to store the keys & ranks
 /proc/jobban_list_for_mob(mob/M)
 	if (!M) return 0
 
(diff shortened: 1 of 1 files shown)


oisin100:
All in theory of course... Theres no real way for me to test this,,, Might just be a problem on the DB side...

oisin100:
Even if this isnt the problem isnull is more efficient then "!="

oisin100:
Adam claims its still connected to the database. perhaps the issue is that when it loops through all the rows and tries to do unbanned != 1 the loop crashes and doesent return any rows.
Honestly dont know jack shit about the way MySQL handles NULL. But this was the only thing changed since then so meh.

DavidEdwards:
The issue is the database was changed, but the new code isnt on the live server yet. The issue will be resolved once the server has the new code. Im waiting for the server to restart at the moment.

PR #410, opened 2016-02-28
Fixes a few SQL problems (Hopefully)

X-TheDark:
Title.

Null checks just to be sure, and according to [Official MySQL reference](http://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#operator_equal), equality is compared using '=' and not '=='.


ghost:
THIS IS CAUSING THE ISSUES WITH THE BANS SPEED MERGE

X-TheDark:
Going to be directly fixed and pushed by Kn0ss0s.

PR #413, opened 2016-02-28
Telescience redesign

Alek2ander:
Redesigned telescience. The telepad can now only target telescience beacons, and access a 9x9 area around them (range=4).
EMPed beacons show up on the telesci computer in the wrong location and can't be precisely targeted.
Telescience beacon and advanced telescience beacon added to the protolathe. The advanced beacon has twice the range (17x17, range=8) and requires diamonds.
The quality of the capacitor in the telepad now affects the delay between uses.
Telescience no longer uses bluespace crystals in any way, except in telepad construction.

Plans for the future:
- Beacons with special functions (Photo beacon, camera beacon, gate beacon, remote-control beacon, something to do with suit sensors)
- Emag interactions
- Other suggestions welcome.


Kmc2000:
I think emagging a console should give the option to overload a beacon, which downs people with an effect similar to the honk blast. 

Otherwise, fuck yes, i can FINALLY play blob without keekeons destroying me

ShadowDeath6:
Remove telesci as a whole :^)

Wardog390:
> Remove telesci as a whole :^)

DavidEdwards:
Lets give this a go.

Since the new TeleSci is on the dev branch, if you are familiar with telescience ways, please give the system a few test runs on your local servers.

EMT321:
The telescience rework does effectively nothing to stop teleport-bombing, whilst making it highly challenging to retrieve bodies via suit sensors. Previously, the only way I could teleport into the AI core was if I either had RD access to get the crystals from dismantling teleporters, or otherwise went on adventures.

As a scientist, I can now teleport into the AI core at roundstart, with the shortest time being about 2-3 minutes. The current solution does not mitigate the issues of telesci-bombing, whilst simultaneously making it nigh-impossible to use telesci for one of its few wholesome ends: retrieving bodies. Additionally, the current implementation is borked, with "camera-beacons" having no specific benefit.

I say all of this from the perspective of a player who recognizes that telescience, in its current form (as of a week ago) is unbalanced. I think taking a page out of other TG codebases, in removing telesci from roundstart and hence putting it behind a barrier of both bluespace crystals and tech levels, is the optimal solution. Take a look at TG and Hippie, both under this system, and you will see a marked lesser number (read: none) of complaints regarding telesci.

PR #419, opened 2016-02-29
Lighting Patch

TehFlaminTaco:
Flashlights now give off light on tables and such again.
"Dims" no-longer build up when turfs are changed.


DavidEdwards:
Forgot - please submit a changelog.

PR #423, opened 2016-03-01
Taco's Changelog

TehFlaminTaco:
Added March First Changelog.


DavidEdwards:
We changed to using the changelog as in the example here:

html\changelog\example.yml

Simply copy that file, and make the necessary changes inside it with notepad.

We don't edit the html directly anymore (as it causes problems down the road)

PR #424, opened 2016-03-01
Adam's Crapple

ghost:
Three major changes:
-rezadone simplified to produce 3u when you mix: 3u cryoxadone, 1u copper, 1u cryptobiolin.
-all hyposprays(including borghypo) can now penetrate hardsuits.
-changeling last resort removed(not entirely removed, it is just commented out).

### DIFF
--- a/code/game/gamemodes/changeling/powers/headcrab.dm
+++ b/code/game/gamemodes/changeling/powers/headcrab.dm
@@ -1,4 +1,4 @@
-/obj/effect/proc_holder/changeling/headcrab
+/*/obj/effect/proc_holder/changeling/headcrab
 	name = "Last Resort"
 	desc = "We sacrifice our current body in a moment of need, stunning and damaging everyone nearby. If a dead body is nearby we infect it to raise again."
 	chemical_cost = 20
@@ -24,4 +24,4 @@
 			crab << "<span class='warning'>You burst out of the remains of your former body in a shower of gore!</span>"
 	user.gib()
 	feedback_add_details("changeling_powers","LR")
-	return 1
+	return 1*/
--- a/code/modules/reagents/Chemistry-Reagents/Medicine-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Medicine-Reagents.dm
@@ -119,7 +119,7 @@
 /datum/reagent/medicine/rezadone
 	name = "Rezadone"
 	id = "rezadone"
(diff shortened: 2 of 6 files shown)


ghost:
Fixing three issue req's in one go?

Good shit

PineappleWolf:
Adam what are you on

ShadowDeath6:
> rezadone easier to make

Xenobio has never been better

Super3222:
In your changelog you put all types of hypo's now can penetrate through hardsuits... but the original one cannot (and shouldn't).

ToGWtF:
@Super3222 

Why? You need to get within touch distance for a hypo to work.

ghost:
No super.

All types of hyposprays are already able to penetrate hardsuits except for the CMO's hypospray, which doesn't make sense.

The CMO's hypo is the only one that is a grand theft obj and yet minihyposprays are better and penetrate hardsuits too. Which is a consistency issue.

Hardsuits being penetrated by hyposprays is intentional and well balanced. If a minihypo can do it, so should the vastly superior borg and CMO hypos.

ShadowDeath6:
> vastly superior 
> Cmo's hypo is literally no different now that it can penetrate hardsuits
> Borg hypos not being the only good hypo

ghost:
Read what I wrote.

Previously, the CMO hyposprays and borg hyposprays could not penetrate hardsuits.

Now, CMO hyposprays and borg hyposprays CAN, just like the minihypo was.

If the minihypo is a miniature hypospray, the grand theft objective hypospray which is apparently more advanced should also be hardsuit-penetrating.

ghost:
With this PR, all hyposprays, including regular ones and borg ones, can penetrate hardsuits. I can give extra penetration to only CMO hyposprays and borg hypos because that is what you're arguing about now.

ShadowDeath6:
If you removed penetration on minihypos it would make miner medics complete garbage on the field. Their entire purpose was to go through hardsuits and its usage was probably an oversight by its original coders.

Super3222:
Oh well, my point was mini hypos were only created to be used and specialized for miners exclusively. So it would definitely make sense if it were able to penetrate hardsuits. I didn't see much of a problem with hypospray's being unable to pierce hardsuits, but if it's consistency problem than it's an okay change.

So the CMO will be an actual threat to nukeops if they are able to fill up their hypospray with neurotoxin. :+1:

PR #455, opened 2016-03-04
Hotfix for indent error causing glitches with other modules

PineappleWolf:
Hotfix as in title to fix errors. MERGEEEE ME


Cruix:
HONK

Issue #431, opened 2016-03-01
Cybermen Feedback required


Cruix:
I have the next round of cybermen updates ready, but before I push I'd like some feedback, particularly from @Alblaka. Here are the changes so far:
*Health analyzers now detect cybermen, and recommend brain surgery.
*<strike>Health scanner huds now detect cybermen, and show that they have a disease.</strike>
*All cybermen abilities have action icons.
*Admins are able to set specific cybermen objectives, including custom ones.
Cybermen Panel updates:
*Admins are able to toggle whether an objective will win the game for the cybermen.
*Admins can edit the progress, cost, and innate processing power of active hacks.
*Admins can remove hack objects from the hacked object list to force cybermen to complete objectives again.
New Hacks:
*All bots, not just securitrons. Emags them.
*Secure crates from cargo. Opens them.
*Microwaves. Does nothing, but was added to the Exploit objective hackable objects list.

I would like some ideas for new hacks, and I was thinking about reducing the cost of converting someone from 200+400_(current cybermen),  to 400+200_(current cybermen). Cybermen are the only antag I know of that takes longer to convert people the more people you already have, and costs get pretty insane late game. The game starts with three cybermen, so the cost of conversion is already the second-most expensive hack in the game, and at seven cybermen, the cost is tied with hacking the AI as the most expensive, assuming no loyalty implants. To discourage murderboning, in an earlier update I put a cap on conversion costs for the "convert or kill non-cybermen" exterminate objective - should I move this cap to always be in place?

ShadowDeath6:
Don't make HUDs detect cybermen, anything that gives a heads-up to antags will always end badly. Health scanners are pretty iffy, but it needs some tell, so they work pretty well.

Add hacking to firing pins to emag them if you haven't already.

pls make cyberman broadcast display names, it makes me cry always because it gives off the feel of being a dev gamemode. Also maybe change the font if possible, make it more like cult chat. 

;______________________________;

Cruix:
> Add hacking to firing pins to emag them if you haven't already.

Do we need this? I have honestly never seen a firing pin that did not work for everyone, except syndicate pins and test range pins.

> pls make cyberman broadcast display names

There was an issue about this before. It has nothing to do with being a dev gamemode, Alblaka decided he didn't want names on it to make it a clunkier method of communication than binary chat. I'd prefer the font to stay distinct from cult since it is pretty much the same color, but I'm open to changing it.

ShadowDeath6:
[...]

PR #467, opened 2016-03-05
Cybermen Updates v4

Cruix:
Refers to issue #431.

Numerous Cybermen updates:
*Changed the cost of Cybermen converting a human from 200+400x(current cybermen) to  400+200x(current cybermen).
*Health analyzers now detect cybermen, and recommend brain surgery.
*Cybermen can now hack secure crates to open them.
*Cybermen can now hack all bots, not just securitrons, to emag them.
*Cybermen can now hack microwaves. Does nothing. Added to objectives as a potential hack target.
*All cyberman abilities now have action icons.
*Admins are now able to set specific cyberman objectives, including custom ones.
*Admins now have much greater ability to change the properties of active hacks and objectives through the cyberman panel.
*Cybermen now leave fingerprints on anything they hack, even if they are wearing gloves.
*Cybermen now automatically focus on any hack they start.


ghost:
Hacked microwaves should just emit tons of radiation when turned on.

COOK FOOD WITH GAMMA RAYS

PR #480, opened 2016-03-06
Borer panel fix

MacHac:
### Intent of Pull Request

+Admins can now 'borerify' players through the player panel.  It only works on humans.

Noticed that the admins didn't have this option, added it.

NOTE: The commits regarding git artifacts are a result of my odd setup, and can be ignored.


code/modules/mob/transform_procs.dm
```diff
@@ -432,6 +432,32 @@
...
+	invisibility = 101
+	for(var/o in organs)
+		qdel(o)
+
+	var/mob/living/simple_animal/borer/B = new /mob/living/simple_animal/borer(loc)
+
+	B.invisibility = 0
+
+	if(mind)
+		mind.transfer_to(B)
```
oisin100, line 454: You might want to use the borer `transfer_personality` proc here so that the borer hud updates correctly.
Can be done with `B.transfer_personality(src.client)`

MacHac:
Noted, thanks for the heads-up.  I was just following the template of the other transform functions and did not notice that was an option.

PR #507, opened 2016-03-08
Restart votes now require +ADMIN instead of +SERVER and more.

AsV9:
This is due to the fact that when only moderators are on the server restart votes could be called.

In addition, I have added a button for mods to click on which toggles restart votes without the need for +SERVER, only +ADMIN. This lets moderators toggle the modes when they are alone on the server. This new button is located in the server panel.

Finally, the change located in cigs_lighters.dm is due to compilation warning. It was also sorta wonky coded. Please let me know if it was coded like this for a reason.

### DIFF
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -531,7 +531,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 		var/hitzone = user.zone_sel.selecting
 		H.suppress_bloodloss(src.stop_bleeding)
 		H.apply_damage(5, BURN, hitzone)
-		if (H.blood_max == 1.5 to INFINITY) // basically, if it's to severe of bleeding damage, a simple lighter won't fix it. maybe try something stronger?
+		if (H.blood_max >= 1.5) // basically, if it's to severe of bleeding damage, a simple lighter won't fix it. maybe try something stronger?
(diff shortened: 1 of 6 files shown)


code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -531,7 +531,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 		var/hitzone = user.zone_sel.selecting
 		H.suppress_bloodloss(src.stop_bleeding)
 		H.apply_damage(5, BURN, hitzone)
-		if (H.blood_max == 1.5 to INFINITY) // basically, if it's to severe of bleeding damage, a simple lighter won't fix it. maybe try something stronger?
```
DavidEdwards, line 534: Nuts. I saw that before too. You are right. Wonky code :D

DavidEdwards:
Just a changelog now chief.

AsV9:
Oh, right, forgot that. Sorry!
Hang on.

code/game/objects/items/weapons/cigs_lighters.dm
AsV9, line 534: Yeah, I got a warning while compiling so I went and fixed it. I suppose I should page @Super3222 considering he coded it, but I'm quite sure he just overcomplicated the code with `1.5 to INFINITY` instead of using the `>=` operator.

Issue #486, opened 2016-03-06
Minebot pull icon overlaps with exosuit storage


Notamaniac:
http://imgur.com/8q5QkAV

Can I also add that their hardsuit helm looks jarring on them, and really needs to be either resprited or made invisible?

ShadowDeath6:
The helmet is an engi helmet, not a miner helmet.

ghost:
I noticed this when I played. I'm fixing it now, but positioning things on the screen is complicated.

Issue #485, opened 2016-03-06
Nerf Minebots


Notamaniac:
Where to even begin...

While the concept of having these is nice, a competent player can easily hop into one of these things, upgrade the entire station, upgrade the ore redemption machine to essentially have infinite minerals and cargo points all within 20 minutes since they have all access. This is all without breaking the drone laws, as you can justify doing this all for performing law 3 as efficiently as possible

Insert exhibits:
http://imgur.com/lHHdO5g
http://imgur.com/y6IYnmQ
http://imgur.com/kUKiQ93

Since this generally shouldn't be able to be done within the first half an hour of the round, as mining is meant to be a 20-30 time gate (at minimum) for science to be able to finish their research, these things really need some sort of nerf so they aren't essentially god tier miners who can also do R&D.

So here's my list of nerfs
1: Remove their ventcrawling and all access so they can't easily waltz into science and do all of R&D. Better yet would be to whitelist the consoles they can interact with to only the shuttle cockpit and the mining shuttle consoles so they can't do R&D or anything else stationwise period

2: Remove their screwdriver repair and force them to use an APC to repair instead (and make sure it's in their activation text so they actually know to use it) so they're forced to return back to the station to repair if they get damaged. I know it doesn't make much sense for an APC to be used, but
using minerals/otherresource would allow them to hoard them and use them at will

3: Remove their drone vision so they can't see where crates are or what is in artifact rooms. This stops them from immediately being able to spot the rooms with sonic jackhammers/diamond drills/ other useful mining stuff.

4: Nerf their speed to borg levels so they can't scarper fifty lightyears away from any threat .I'm a bit iffy on this one, in general mining mobs need a buff as they're no threat at all anymore unless you encounter two or more of them at the same time, but I'm digressing. This'd also slow down the rate they can deliver minerals, which is a good thing in the long run.

5: Give them the normal scanner instead of the advanced one. The advanced scanner simply lets you spot minerals more consistently and let's you find diamonds and uranium much more easily, while most miners have to at least put a token effort in mining first before they can get it.

ToGWtF:
I thought the plan for them was to not allow them to leave the mining asteroid for any reason?

ShadowDeath6:
Why is their effectiveness a bad thing? They were made specifically because miners would never bring in materials, completely cockblocking science from upgrading things. Their laws allow them to do R&D for better mining supplies, but not for upgrading things on their own.

Super3222:
[...]

PR #510, opened 2016-03-09
Minedrone Nerfs

Cruix:
Committed for Groxic because his Git isn't working.
Addresses #485 and #486

Minedrones have been nerf'd and given a mining points card. Their HUD is fixed to prevent overlapping of drop and secondary storage icons.


ShadowDeath6:
> not starting with kinetic accel

Notamaniac:
:+1:  They still get the advanced scanner and they're stuck with the engi helm, but eh,

PR #513, opened 2016-03-09
Changes the duct tape sprite to look less scarf like

Notamaniac:
Figured I'd try to learn to learn git to at least do basic things
This PR should replace the duct tape scarf with this
ducttapeeg
If I managed to actually blow anything up by just changing the dmi file and adding a changelog, colour me impressed and yell at me.


AsV9:
:100: \* :+1:

PR #527, opened 2016-03-11
Nerfs Morgue Prisons

Cruix:

Makes it so you can move out of lockers and bodybags while they are closed inside of a morgue. This prevents morgues from being used as an inescapable prison.


Cruix:
Oddly, you could resist out of welded lockers that were inside morgues, but you could not move out of unwelded ones.

PR #536, opened 2016-03-12
Critical borer updates.

oisin100:
This need to be merged ASAP.

I fucked up big time with the transformation_procs resulting in anyone using a transformation proc who doesn't have a borer in them(EG, Lings, Geneticists) Would get kicked out of their body by a runtime.

Also fixes borer jobbans.


Ephemeralis:
Merged.

PR #540, opened 2016-03-12
Duct tape boogaloo - Non massive edition

Notamaniac:
It turns out that you really should look at the scale of other sprites before actually making them. And that AA on small sprites don't work out that well.

Old
ducttapeg
New
ducttape2


Ephemeralis:
Merged.

PR #541, opened 2016-03-12
Re-adds the McDonald costume to the AutoDrobe's normal selection

Super3222:

Removes the mascot costume from the AutoDrobe's premium list.
Re-adds the mascot costume to the AutoDrobe's normal list of costumes.
This will also change the costumes description slightly.


Ephemeralis:
Proud of u.

Issue #448, opened 2016-03-03
Make headslug bursting make the slug monkey a changeling (Xenobio)


ShadowDeath6:
### Expected Behaviour

Monkey created from headslugs via xenobio gets changeling antag status and powers. 
### Actual Behaviour

Monkey bursts from corpse but has no antag status, and by extension, no powers. 
### Steps to reproduce the problem
1. Spawn headsluggo from gold slime extract. 
2. Use sentience potion on headsluggo. 
3. Headslug implants an egg in a human corpse. 
4. Monkey pops out like 5 minutes later
5. Monkey is sentient and can be turned human via genetics
6. Monkey has no powers. 
7. Bitch in deadchat about not having infinite ling army. 

The reason for this is that headslugs aren't inherently changeling antags, and since lings get their powers from their antag status, they don't get powers.

Possible fix- force antag status onto people who become the headslug mob.

oisin100:
Headcrabs no longer exist in the latest updates.

oisin100:
Easy fix. Headcrabs have a separate mind holder for the actual ling mind `var/datum/mind/origin`

Time-Green:
I am not a fan of making antags.

I AM a fan of making lingcurity

FluffySurvivor:
LING FACTORY? 
GIMME

ghost:
/tg/ has this and it's not unbalanced.

Interestingly, the changeling still has to obey their master there and basically it's just a ling forced to serve you.

It's quite fun and it doesn't make an antagonist per se, just grants the powers. Without objs they can't do anything their master doesn't want them to.

ToGWtF:
@oisin100 
teach meeee

ShadowDeath6:
PR when

oisin100:
Ill do it then...

oisin100:
Fixed in #550

PR #550, opened 2016-03-12
Fixes pink slime extracts used on headcrabs not working right.

oisin100:
Ability for Xenobio to create Changelings. Yey.
These changelings do not have objectives and arent given antag status. They are normal crew and should act and be treated like normal crew.

Theres 1 runtime error what occurs but im pretty sure its not related to this change but ill add "Needs Testing" anyway.

Implements: #448

### DIFF
--- a/code/game/gamemodes/changeling/evolution_menu.dm
+++ b/code/game/gamemodes/changeling/evolution_menu.dm
@@ -351,7 +351,7 @@ var/list/sting_paths
 		user << "<span class='danger'>You lack the power to readapt your evolutions!</span>"
 		return 0
 
-/mob/proc/make_changeling()
+/mob/proc/make_changeling(var/createhuman = 0)
 	if(!mind)
 		return
 	if(!ishuman(src) && !ismonkey(src))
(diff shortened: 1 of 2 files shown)


DavidEdwards:
I don't know how I feel about adding changelings like that. People hate them enough as it is.

ShadowDeath6:
it's not adding hostile changelings, just making ones that are loyal to the xenobio.

ghost:
it works fine on /tg/ and it's balanced there

antagless changelings are the least OP things xeno can make

this actually takes effort

statues are unkillable murdermachines that don't take effort

DavidEdwards:
Two things I worry about, if you can convince me with these two we can merge:
1. Our rules mean people don't know what lings are. Is this an issue if lings can be spawned. Should this mean a rule change?
2. Will this result in chaos? People screaming they see an alien. Perhaps a good thing, if it causes distractions for antags.

AsV9:
I'm fine with this.

oisin100:
> Will this result in chaos? People screaming they see an alien. Perhaps a good thing, if it causes distractions for antags.

Im pretty sure Xenobiology already does this.

> Our rules mean people don't know what lings are. Is this an issue if lings can be spawned. Should this mean a rule change?

That's true. But xenobiologists already know what Xenomorphs and borers are and actively aim to get them

ghost:
Kn0ss people know what lings are.

ShadowDeath6:
:+1:

PR #552, opened 2016-03-13
Fixes 'Toggle Antag Icons' Not showing gang members correctly.

### LINKED ISSUE #450: Admin antag HUD doesn't work with gangs
### Expected Behaviour

Admin antag HUD shows different colored icons for different gangs, and can distinguish bosses and normal gangsters
### Actual Behaviour

All gangsters show the same [?] unknown antag icon

oisin100:
Just makes life simple on the Admins.

Fixes #450


Alek2ander:
This will give the same icon to all gangs, different gangs should really have different icons.

oisin100:
> This will give the same icon to all gangs, different gangs should really have different icons.

This will require some Spriter support. Does someone want to make the sprites?

Time-Green:
Wait, if you don't give every gang a special color. Then this was pointless

oisin100:
> Wait, if you don't give every gang a special color. Then this was pointless

It shows if they are part of a gang and if they are a gang boss. Where as before it would display them as a unknown antag. Ill see what i can do to distinguish between gangs.

oisin100:
I came up with a much better way to do this.

PR #555, opened 2016-03-13
Fixes limited number of role preference bitfields

### LINKED ISSUE #544: Update preferences to avoid bitfields
### Expected Behaviour

Make a game mode, allow users to play it.
### Actual Behaviour

Make a game mode, nobody can play it because Byond has shitty integers. 
1. Make gamemode
2. Cry when you exceed the allowed size of an integer

Cruix:
Fixes #544.

I know Kn0ss0s is already working on this, but I was struck by a moment of inspiration and made this. It replaces the be_special var with a list of integers, and expands this list if there are more preferences than can fit in the space it has currently. To facilitate this the BE_\* bitflags in preferences.dm have been changed from bitfields to ascending values starting at 0; they are converted to bitfields with the formula bitfield = 2^(pref%16), with floor(pref/16) determining which index of the list to use. I have tested it in passing, and it appears to work, but I would not be surprised in the slightest if I missed something.

I also fixed some procs in code/modules/client/preferences.dm that were defined in a nested way, which is against the /tg/ coding standard. This involved shifting a lot of indents, which is where 90% of the diffs come from. The most important diffs in that file are on lines 143-153.

I also enabled Cyberman and Zombie prefs, and make the Revenant pref visible in the lobby selection menu.


oisin100:
Interesting... The code looks great!

DavidEdwards:
Ill upload my code to a branch so you can see what I have done. I added an improved selection window, options condensed into a single style. You'll see what I mean when I push it.

PR #568, opened 2016-03-14
Tactical animal nerfs

Cruix:
Refer to issue #0.

*Tactical dolphin heath reduced from 150 to 100.
*Tactical carp health reduced from 300 to 150.
*Tactical animal fire rate reverted from 1 salvo per two seconds to 1 salvo per second.
*Tactical animals now have a recharging battery in their harness that is drained by shooting. The maximum charge is 1000, a shot costs 150 for a carp and a 3-round-burst costs 225 for a dolphin, and the battery charges at a rate of 10 per second. These values are subject to change. Charge can be seen by the animal in its Status panel.


ShadowDeath6:
Now they suck >:^(

PR #572, opened 2016-03-14
Orbit fixes

AsV9:
Ported tgstation/-tg-station#13702
- Makes orbiting much less reliant on `animate()`.
- All orbits now start at North instead of East

`SpinAnimation()` has had it's logic un-hardcoded, that is to say, it no longer forces 3 segments of rotation on Atoms, it instead contains an argument allowing potentially infinite segments (I suspect post 360 to get a bit ~weird~)
Profiling reveals that this performs at the same CPU cost.

This should help delag the tesla.

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -252,3 +252,10 @@
 #define NO_RESULT				-4
 #define QUERY_OK				-5
 #define INSUFFICIENT_CREDITS	-6
+
+//Ghost orbit types:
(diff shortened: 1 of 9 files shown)


ToGWtF:
please no orbit. Its annoying and cluttersome. I would rather just a static ghost, so I can actually see the player under it

ShadowDeath6:
orbit is super annoying, can we just remove the tesla to stop lag?

AsV9:
No more ghost orbiting. Fixes are still in though.

Issue #592, opened 2016-03-16
You can breathe any gas as if it were oxygen


Cruix:
Reported by Adam.
### Expected Behaviour

You die of oxyloss when you are breathing from air that does not have enough oxygen in it.
### Actual Behaviour

The checks for species that breathe non-oxygen gasses think that the fact you don't breathe those gasses means you are able to breathe where you are.
### Steps to reproduce the problem

Spawn in as atmos tech.
Fill a tank with pure N2.
Don your hardsuit and jump into space.
Breathe forever because you don't consume n2 but you also don't take oxy damage.

ghost:
yay merged

PR #593, opened 2016-03-16
Fixes breathing non-oxygen gasses as if they were oxygen

Cruix:
Fixes issue #592

Fixes humans not taking oxy damage if they were breathing a gas mix with no oxygen in it.


oisin100:
Isnt this intended for N2?

Cruix:
The bug occurs with plasma and CO2 as well. The reason N2 is used as an example is because there are no negative effects to breathing it.

Notamaniac:
Now people will ACTUALLY have to use internals when they're in low O2 places? How could you?

PR #601, opened 2016-03-17
Adds Medihover borg sprite

utahclock:
Adds V.2 medihover borg sprite


Notamaniac:
The eye sprite needs to be a separate icon rather than drawn on the borg. Also the outline sprite is a bit too purple for my taste, I'd personally darken it to blue but that's my opinion.

The reduced rocking looks a lot better though.

utahclock:
where can i find the borg eye file then. i dont see it anywhere....

Notamaniac:
Just make the icon in the robots file, with it as a normal movie and label it eyes-medihover so the coders know what it is. It might help to overlay the normal borg with a layer using GIMP or a program with layers so you can easily do it for the animation states without having to check back and fro

utahclock:
right found the other eyes in the dmi dont know how i missed those, eyes are done now. Ill also need a robo transform too. Ill work on that now and push version 3 soon. with bluer outline too.

utahclock:
closing for version 3.

PR #603, opened 2016-03-17
Fixes borer hud breaking antag huds such as gang.

oisin100:
Updates borer hud so that it doesent break antag huds such as gang.

Also fixes the "Make Borer" admin verb so that it gives the borer the borer message + its objectives


DavidEdwards:
Changelog, honk.

oisin100:
Wut. Not sure what happened with this..

PR #608, opened 2016-03-18
Update README.md to reflect recent updates.

oisin100:
Just so we dont get asked "WHY CODE NO COMPILE?!?!"


DavidEdwards:
Good catch!

PR #615, opened 2016-03-18
Makes borers impervious to bombs and other stuff

oisin100:
Makes borers impervious to explosions whilst inside a host.
Also broadcasts borer hivemind to the ghosts.


timkoster1:
Even more borer buffs?
Cmon dude
At least at some easy way to remove them
When you remove them with surgery they just leave, and take your body instead

oisin100:
> When you remove them with surgery they just leave, and take your body instead

Ill make borers fall asleep when removed with surgery. Or just kill them outright when removed.

ShadowDeath6:
Making borers die immediately upon removal just makes it all the more reason for powergamer mckeek to run to surgery instead of trying to reason with the borer, though, since the borer has literally no chance to live upon removal.

PR #125, opened 2016-01-23
Traitor Jetpack obj

Groudonmaster:
-Adds new Jetpack subtype
-Special icon state specifically for the Captain's Jetpack
-Reupdated Traitor objective items to be for that specific jetpack
-Other commits are me sloppily using git to move them to one branch.

PR #625, opened 2016-03-19
Gold Jetpack now appears on your back

Super3222:

Adds the gold jetpack icons to back.dmi, so it appears on your back.
#125 didn't do it, and it's been invisible since. And that's a BAD THING, considering it's a traitor objective and they could just run around with it on undetected.


Groudonmaster:
I forgot to do this? I swore I did. Maybe I messed it up when I combined everything into one branch. O well thanks Super

PR #634, opened 2016-03-20
Deployable Barrier Buff v1

Super3222:

Currently deployable barriers are the easiest thing to get through, even without an emag.

Currently you could destroy a barrier with a toolbox or a fire extinguisher under 10 seconds and it goes completely unnoticed until someone walks over and sees the rods left over from the barrier and the small environment damage.

So my mission is to make it loud enough for someone from a long, but fair distance to hear it.

I could use feedback from people who play security often.

### DIFF
--- a/html/changelogs/Super3222-deployablebuff.yml
+++ b/html/changelogs/Super3222-deployablebuff.yml
@@ -1,0 +1,13 @@
+# Your name.  
+author: N3X15
+
+# Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+
+# Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
Make it explode and damage the person who broke it

Time-Green:
Make it send a message on security channel where and who destroyed it.

EMT321:
Making them explode seems like a beneficial way to stop people smacking em open.

Currently, most people dont even try.

DavidEdwards:
This PR is small enough to be put in without too much fuss for balance. Lets see how it goes.

html/changelogs/Super3222-deployablebuff.yml
```diff
@@ -0,0 +1,13 @@
+# Your name.  
+author: N3X15
```
Cruix, line 2: Might want to update this.

PR #643, opened 2016-03-21
Cybermen Updates

Cruix:

*Added some logging to buttons in the cybermen panel.
*Hacked AIs and cyborgs are now added to the cybermen antag hud. This is not revoked if the silicon is de-converted.
*Finally fixed the "Hack AI" objective. Actually fixed this time.
*Admins can now queue a cybermen objective and see if it is valid before assigning it.
*Fixed a grammatical error in hacking humans that are dead or braindead.
*Fixed a runtime when a cyberman tried to convert a catatonic human.
*Cybermen now get a data hud that appears over objects being hacked. Can't wait for this to break horribly.


ShadowDeath6:
> hud not removed upon de-conversion

Would that mean that an AI would still need to obey them, or would it be used for validhunting?

Cruix:
It would be used for validhunting. I added the hud to reduce confusion for the silicon over who was a cyberman, and in order to remove it when the silicon is deconverted, I would have to add a specific check whenever the silicon's law 0 was changed, which I would prefer not to do. It is extremely unlikely that this will happen in a normal cybermen round, since there will be no emags or hacked modules (Is there a way to remove ion and zero-laws without traitor gear?), but I can make it go away if it turns out to be a big deal.

Just to be clear, the hud is still removed for humans who are deconverted, it is only cyborgs and AIs who keep it.

ShadowDeath6:
Cybermen 2: Revenge of the AI

FluffySurvivor:
Can we add names to who sends messages in the cyberman chat? Currently it's a messy clusterfuck as you don't know who says what.

ShadowDeath6:
I already made an issue for that and it got shot down, Fluffy.

Super3222:
It's meant to be like that and not another group antag mode where communication is 100% easy.

FluffySurvivor:
It's not only hard, but it's FRUSTRATING. 
It wouldn't be such big of a deal if it were just inconvenient to use, but the current shit hurts my brain. 

A group antag that doesn't promote communication? That's garbage. 

Lings which are SOLO antags, get ultra easy, built-in, stealthy,  guaranteed, usable-in-crit and instant team communication but defined GROUP ANTAGS have to struggle with some messy text that confuses the shit out of you. 

Rip balance.

PR #645, opened 2016-03-21
Fixes toggleable energy guns shooting the wrong type of shot once after toggling

### LINKED ISSUE #444: Hybrid tasers only switch after firing once after switching
### Expected Behaviour

Switch to disabler from taser, shoot disabler
### Actual Behaviour

First shot is a taser instead of a disabler, fixes itself after the first shot. Still annoying.
1. Get Hybrid taser
2. Switch modes
3. Shoot

Cruix:
Fixes #444

Fixes toggleable energy guns shooting the wrong type of shot once after toggling. The old shot is refunded.

### DIFF
--- a/code/modules/projectiles/guns/energy.dm
+++ b/code/modules/projectiles/guns/energy.dm
@@ -58,7 +58,7 @@
 	if(chambered && !chambered.BB) //if BB is null, i.e the shot has been fired...
 		var/obj/item/ammo_casing/energy/shot = chambered
 		power_supply.use(shot.e_cost)//... drain the power_supply cell
-		chambered = null 
+		chambered = null
 		newshot()
 	else
 		chambered = null //either way, released the prepared shot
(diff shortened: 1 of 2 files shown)


ShadowDeath6:
Will this break syndieborg LMGs?

Cruix:
Don't think so, I'll test it. How exactly would it break them?

ShadowDeath6:
The LMG is an energy weapon that uses bullets, so it needs to have bullets in the 'chamber', which had to apply to other energy weapons because it's a subtype. Looking through the code, it shouldn't have any conflicts since the actual ammo stays consistent in the LMG.

EMT321:
Thank you, no more accidental lethals hopefully..

AsV9:
Did you test it, @Cruix?

Cruix:
I did, there were no issues I could find.

PR #666, opened 2016-03-23
Fixes missing tiles from CentComm.

### LINKED ISSUE #595: CentCom missing floor tiles
### Expected Behaviour

Mappers not break things.
### Actual Behaviour

Human nature occurs.

AsV9:
Fixes #595 
Also #666 get.


ShadowDeath6:
Wow removing perfectly good map features
Why do chefs deserve the right to not be near exposed space?

Issue #642, opened 2016-03-21
Overhaul antag selection system


ParagonPrime:
Right,

As we're all aware, the ratios for people being selected for antag is... really skewed. While some people end up antag more often than not (according to Adam, one player has a chance of about **75 percent** of being antag), others basically never seem to get it.

So! Let's overhaul RNGesus a bit and make it more fair. Because let's face it, there's really no logical argument against people having a RELATIVELY FAIR chance of being antag.

Basically, let's make it so that, the more often you're antag, the lower your chances of being antag become- so if you roll lucky twice, you've got a lowered chance on the third time. And vice-versa; if you've had a long streak of being nonantag, you'll have a higher chance of being an antag in subsequent rounds. This also applies within antags themselves; if you've gotten one type of antag a lot, your chances for other types goes up.

Should be relatively simple and is fair to everyone.

DavidEdwards:
The system should be entirely random.

The players list is shuffled / randomized (I think twice even) during the selection process.

I would be open to a check based on previous round(s) "luck". This will involve database work, so I should probably take this one.

Time-Green:
This seems like a really cool idea. And it makes variations more common.

I like it, alot

FluffySurvivor:
I disagree with having reduced chance of antagging in subsequent rounds. That would create really weird patterns of antagfishing, so no. 

Admins should never interfere with things such like antag selection, admins will abuse this to make sure certain players never get antag.

Time-Green:
> increases fluffies cult chance to 100%

DavidEdwards:
I certainly won't let admins change the values. 

It would be a weighted system. 

Whereby:

Each time you antag, decreases the weight. 
Each time you don't get antag, increases the weight.

Capped to a min and max.

It really doesn't need to be more complicated than that.

ParagonPrime:
Okay, let's remove the admin-bus bit entirely, then. DavidEdwards seems to have it, though the min/max should scale with server population/number of antags in the round a bit.

Edited.

Also, Fluff- what's your reasoning? "Weird antagfishing patterns"? All this does is make it so that you can't be the tator three, even four rounds in a row, as I've witnessed with some people. Not even you, necessarily.

FluffySurvivor:
> increase fluffies cult chance to 100%
> uninstall

X-TheDark:
Also, how would this work with having the preference off/quiet rounds? Weight is not changed at all?

[...]

PR #672, opened 2016-03-24
Antag Weighting System

DavidEdwards:
Refer to issue #642.

In order to help ensure that everyone gets a turn to be antag, we have gone less random and more planned in antag selection.

I need a sanity check on the code in this PR. Comments are provided.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -291,6 +291,58 @@
 	if(security_level < SEC_LEVEL_BLUE)
 		set_security_level(SEC_LEVEL_BLUE)
 
+/datum/game_mode/proc/pick_candidate(role)
+	if(dbcon.IsConnected())
+		var/list/ckeys_to_drafted = list()
+		var/ckeys_drafted = ""
+
+		for (var/datum/mind/player in antag_candidates)
+			ckeys_to_drafted += get_ckey(player)
+
+		ckeys_drafted = list2string(ckeys_to_drafted, "', '")
+
+		var/DBQuery/query_whitelist = dbcon.NewQuery("SELECT `ckey`, `antag_weight` FROM [format_table_name("player")] WHERE `ckey` IN ('[ckeys_drafted]')")
+
+		if(!query_whitelist.Execute())
+			return 0
+
+		var/list/output = list()
+
+		var/total = 0
+		while(query_whitelist.NextRow())
+			var/ckey = query_whitelist.item[1]
+			var/weight = text2num(query_whitelist.item[2])
+			output[ckey] = weight
+			total += weight
+
+		var/R = rand(0, total)
+
+		var/cumulativeWeight = 0
+		var/datum/mind/final_candidate
+		for(var/ckey in output)
+			var/weight = output[ckey]
+			cumulativeWeight += weight
+
+			if(R <= cumulativeWeight && !final_candidate)
+				weight = max(25, weight / 1.5)
+				var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[ckey]'")
+				query.Execute()
+
+				for(var/datum/mind/candidate in antag_candidates)
+					if(lowertext(get_ckey(candidate)) == lowertext(ckey))
+						final_candidate = candidate
+						break
+			else
+				weight = min(400, weight * 1.5)
+				var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[ckey]'")
+				query.Execute()
+
+		if(!final_candidate)
+			return pick(antag_candidates)
+		else
+			return final_candidate
+	else
+		return pick(antag_candidates)
 
 /datum/game_mode/proc/get_players_for_role(role)
 	var/list/players = list()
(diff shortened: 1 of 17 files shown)


Time-Green:
Nice, I only get antag every 7 days while mr Celia DeRose got a 4 antag streak.

I love this idea. What would happen if a person observed?

DavidEdwards:
Only people who are part of the antag selection pool would actually have their weights altered. 

So if you aren't taking part. Have quiet mode on. Don't have antag preferences enabled. Then you won't be changed.

Not sure what you mean by if a person observed.

Incidentally, this PR isn't complete. I would need to change the `pick(antag_candidates)` function call in every game mode. I left it for now, to keep the analysis simple.

Super3222:
Excellent...

DavidEdwards:
Seriously though @yogstation13/coders @yogstation13/senior-coders 

This needs holes poking in it.

DavidEdwards:
`AdamElTablawy self-assigned this 31 minutes ago`

Does that mean you're going to poke some holes?

AsV9:
The code **seems** good to me.

ShadowDeath6:
Breddy gud :^D

code/game/gamemodes/game_mode.dm
```diff
@@ -291,6 +291,72 @@
...
+		// Find a number between 0 and our weight upper bound
+		var/R = rand(0, total)
+
+		var/cumulativeWeight = 0
+		var/datum/mind/final_candidate
+		// We will loop through each antag candidate until we find the point where the
+		// random number given intersects with a candidate. All other candidates will
+		// have their chance to be antag increased while the selected candidate will have
+		// their chance decreased
+		for(var/ckey in output)
```
X-TheDark, line 333: **EDIT: Nevermind, I don't ~~know how it works~~ trust it. This system/code would work perfectly fine. Post below not removed for clarification (and my humiliation) sake.**

Shouldn't the list of eligible ckeys be randomized/shuffled before being iterated through, as depending on how the database sorts these records, you might still end up with a skewered system.

What I mean by this is that if we take, for example this trivial situation of having 4 players with the following weights:
aaaa - 25
bbbb - 26
cccc - 27
dddd - 28

The first 3 players will be preffered for the antag selection (because of the order they are looked at) before the logical 4th player with the highest weight being selected. Although the system will equalize (since weights are decreased/increased), it might take quite a long time if we are talking about something like a 50 player server where everyone has antag preference on.

DavidEdwards, line 333: The odds aren't really in favour of any of them at that level. The difference is so minimal. It is essentially 1/4 chance there. Basically what would happen in that instance:
1. All the weights are added together: 106
2. A random number is generated. Lets say 95.
3. We loop through the list from the start.
4. We add the first players weight to the cumulative total (now 25).
5. Is 95 less than the cumulative total (25)? No. Go to the next player.
6. We add the second players weight to the cumulative total (now 51).
7. Is 95 less than the cumulative total (51)? No. Go to the next player.
8. We add the third players weight to the cumulative total (now 78).
9. Is 95 less than the cumulative total (78)? No. Go to the next player.
10. We add the fourth players weight to the cumulative total (now 106).
11. Is 95 less than the cumulative total (106)? Bingo. We have a winner.
12. 4th player is taken as the selected player.
13. We continue to the end of the antag list because we need to set the new weights in the database for the other players too.

oisin100:
Wonderfull...

Iamgoofball:
> ckeys_drafted

this isn't a silly variable

drafting is literally the term for this dipshit

DavidEdwards:
I won't be including this in the next update. There's enough going on in that one. But I will be merging this for the next one.

So please take a last look over the logic.

> drafting is literally the term for this dipshit

Love you too goof

Iamgoofball:
<3=====D

TehFlaminTaco:
David, I need my rank here re-added to Coders by the way, I'm still Initiate Here.

DavidEdwards:
Ah sure, Ill take care of that now. Incidentally, don't use Github for that. Please tell me on Discord.

ParagonPrime:
It seems to work in sort of waves; at first I had a total dry spell of antag then got sling, ling, and tator three rounds in a row.

Seems to work absolutely great for just about everyone from what I've seen and heard, though.

DavidEdwards:
Good news, thanks for the feedback.

rweber89:
This doesn´t work. I haven´t gotten antag in days. Closing in on having played 40 consecutive rounds of not being antag.

Instead of weights, I would suggest a system where you count the amount of rounds since every player has last been antag. The player with the most rounds since their last antag assignment will be selected.

That would be the basic system. An advanced version of this could also include antag-specific roles in the counting system, that way everyone gets a fair chance of rotating between antag roles.

And I´m not the only person for whom it is broken: The same guy has just been nuke ops 3 times in a row today, within a few rounds. So both extremes are there. A system where it counts who has been what kind of antag how many rounds ago will create a balanced and fair system.

DavidEdwards:
Let me know your ckey and I will investigate. (and ideally the other guy)

rweber89:
You´ve got mail David.

ParagonPrime:
Yep, the system definetely seems to have a preference for giving dry spells followed by bursts of antagging. I've been nukeops thrice today, twice consecutively.

Which I don't think is necessarily a bad thing. Giving long breaks from antagging followed by a lot of antagging (and uncommon antagging- I've been sling twice and wizard once recently, the former being one I only had once many months ago and the latter absolutely never) keeps it a bit fresh.

Does the weights system also apply within antags themselves (so if you've been tator a lot but never bleb, does bleb get higher chance) or is it just for the title of antag as an overall?

Issue #617, opened 2016-03-18
Let ghosts enter empty soulstones


ShadowDeath6:
Title. Lets ghosts enter cultist soulstones similarly to posibrains, which in turn lets the ghosts play as constructs.

Additionally, let players ghost in and out of soulstones so they can chill out in deadchat instead of either ghosting which permanently fucks them, or suffering in a stone for all eternity. Cultists could not use soulstones without an active person in them to make constructs, obviously.

ToGWtF:
Eh. Letting them ghost back and forth is meh since you can just keep a soulstone in your pocket to feed you info,  but you can already astral journey anyway so ehhhhh

ShadowDeath6:
obviously it would be treated the same as it is for posibrains (metagame).

Super3222:
Hmm, I've made up my mind about a few things. I got a plan for this, so I guess I'm claiming it.

On a side note, I don't think ghosts entering soul stones is a good idea. If a cultist WANTS someone to become a construct, they crit/kill them and than hit em with it. Not leave it out in the open like a drone / posibrain.

Greatbigben:
will close when merged

ShadowDeath6:
The ghost wouldn't be able to inhabit the stone without them leaving an empty one on the ground a la droneshell

PR #690, opened 2016-03-26
Shades can return to their soulstones

Super3222:
#617

Ghosted Shades can re-enter their soulstone homes by clicking on the shard.

When you summon shades out of their soulstones, it'll force the ghost into the shade as well.

FAQ:
'what if i don't wanna play'

The only solution is suicide.

### DIFF
--- a/code/game/gamemodes/wizard/soulstone.dm
+++ b/code/game/gamemodes/wizard/soulstone.dm
@@ -73,6 +73,11 @@
 
 		if ("Summon")
 			for(var/mob/living/simple_animal/shade/A in src)
+				if(!A.key)
+					for(var/mob/dead/observer/G in player_list)
+						if(G.name == A.name && G.name == A.real_name)
+							A.key = G.key
+							G.mind.transfer_to(A)
 				A.status_flags &= ~GODMODE
 				A.canmove = 1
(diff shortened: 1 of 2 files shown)


ToGWtF:
So infinite ghosting to find targets? eh. cultists can already astral journey and ghost anyway

Super3222:
It'll be considered meta if they use any information that they gather from ghosting.

This is just so that they aren't completely fucked if they ghost because of boredom.

DavidEdwards:
I think, if they ghost out of boredom, they get what they deserve.

I'm wary of giving players more ways to meta.

Open to thoughts.

ShadowDeath6:
Reminder that posibrains work the same way and we haven't had any real problems with it

ShadowDeath6:
Merge reeeeeee

Super3222:
reeeeeeeeeeeee

DavidEdwards:
> Reminder that posibrains work the same way and we haven't had any real problems with it

Can you pop in and out of ghost at will multiple times with Posibrain?

ShadowDeath6:
Yes you can, but somebody can take your body while you're out.

ShadowDeath6:
Merj pls ;_;

PR #693, opened 2016-03-26
Fixes fulltile windows at derelict

Notamaniac:
A lot of the fulltile windows on the derelict had random directions set to them, making it seem that you could walk through them, when in reality it was pushed in a direction. My guess was that some coder did a good old ctrl f replace with windows, and didn't pay attention to the direction var, but anyways this fixes that.


Super3222:
The K13 doesn't exist! Stop searching for it!

Notamaniac:
The best way to hide something is to advertise the fact that it doesn't exist.

Issue #67, opened 2016-01-17
Head role whitelist


george99g:
Add an Ultra preference for head roles to people who have been accepted into a whitelist.

killerx09:
(Also AI)

BadminBen:
gonna need to work out the white list first

george99g:
We kind of need the code to put the whitelist into, before we get the actual whitelist.
Everything will be normal even if nobody is on it.

Hydre:
Eeeh didn't we once have heads jobs whitelisted?

Could've been a different server but I don't really think this is a good idea.

EDIT: Unless you mean by ultra preference that whitelisted just get priority over non whitelisted

george99g:
> EDIT: Unless you mean by ultra preference that whitelisted just get priority over non whitelisted

Exactly!

Hydre:
Oh well that's a good idea

BadminBen:
is this being worked on?

DavidEdwards:
It is quite high on my list - unless someone else takes it over. I am
working on Predators right now - since I have been promising Kmc I'd work
on that for ages.

PR #699, opened 2016-03-26
Global Whitelist system for head positions + AI

X-TheDark:
Refer to issue #67

Adds support for whitelisting people for head roles + AI. Whitelisted people get an Ultra preference (higher priority than High). **At the moment, this is a global system, which means if you are whitelisted, you get Ultra preference for ALL eligible positions.**
In addition, minor changes to the UI - Occupation Preferences have a black button on positions that are whitelisted (only visible to people who actually are whitelisted).

New client var : is_whitelisted - cached from DB when client connects, accessed by...
New mob proc : is_whitelisted(mob/user) - retrieves cached value of is_whitelisted

Stuff tested:
- People in whitelist being able to select "Ultra"
- People who are no longer in the whitelist get their "Ultra" preferences removed
- Being able to join the game and be selected for the position you have "Ultra" preference on
- ...more stuff will be added as I test whether the code is working as it's supposed to

UI Screenshot (only seen by people who are whitelisted, non-whitelisted UI unchanged)
untitled


X-TheDark:
If necessary, system can be converted to a per-job basis later.

code/__HELPERS/mobs.dm
```diff
@@ -133,6 +133,16 @@ var/global/list/roundstart_species[0]
...
 		return 0
 
+/proc/is_whitelisted(mob/user)
+	if(ismob(user))
+		if(user.client)
+			return (user.client.is_whitelisted)
+	else if(istype(user, /client))
+		var/client/C = user
+		if(C)
+			return (C.is_whitelisted)
```
oisin100, line 143: Just a minor thing. Are the parentheses  and the if(C) check really required? I think istype does this automatically.

ShadowDeath6:
Tfw you need to whitelist for AI

cool though, hopefully it stems the tide of stupidity in head roles. :+1:

Alek2ander:
It's not like the same metafriends take these roles every time or anything

code/__HELPERS/mobs.dm
X-TheDark, line 143: I'll be honest, I just copy-pasted the is_donator code and rewrote that. As for the parentheses, they are not necessary, I just seem to be able to read that notation more easily.

Also, with istype(), yes, you are right, the if(C) check is not necessary, as it would fail if user is not a /client and if it is, the type-casting is expected to succeed.

PR #714, opened 2016-03-27
Borer Chem Panel

DavidEdwards:
I found the chemical selection verb to be worth upgrading.

I made a full HTML panel to display the chems and their usage.


code/modules/mob/living/simple_animal/borer/borer_chemicals.dm
```diff
@@ -1,5 +1,6 @@
 /datum/borer_chem
 	var/chemname
+	var/chem_desc = "This ia a chemical"
```
ShadowDeath6: "ia"

DavidEdwards: Intentional....... Honest.

DavidEdwards:
honk

Issue #705, opened 2016-03-27
Re-add abduction to mode rotation


MacHac:
### Issue

Abduction, while being sound in theory and fun in practice, was originally removed because the crew had little real way to combat the alien menace and, in fact, rarely saw the abductors at all if they knew what they were doing.  This led to abductor rounds looking very much like extended rounds, at least until they got overrun with horrifically OP spiders.
### Proposed solution

I propose that Abduction be added back into gamemode rotation, but as a secondary mode like changeling.  This would mean that abductor teams could only spawn alongside traitors and possibly changelings.  This would fix the problem of abductor rounds being largely uneventful.  Additionally, the worst glands (bodysnatcher, spiderling) would be removed and possibly replaced by more benign varieties.

ShadowDeath6:
I wouldn't like it to be as common as a gamemode, MAYBE have it be like a 10% chance to spawn in regular traitor ONLY, so we don't have all the fuckery with lings that happens right now.

Greatbigben:
it should be more like 25% its a great addition to the round and makes interesting side effects to the ones that get abducted

Time-Green:
It's alot of fun.
Aaaanndddd. Please fix the abductor monitor, it needs alot of time to load, enough to get your agent killed.

Greatbigben:
i forgot that the monitor was bionked

ShadowDeath6:
I for one find abductor mode one of the least fun modes since you can literally be fucked over permanently or killed because of an abductor gland. If it appeared every 4 traitor rounds we would probably end up with a lot of rounds ruined because the objectives are either retarded or the glands are so annoying you just kill the abductee. If you could remove the gland it would bw fine, but you can't, so you might as well kill yourself because you stun for 5 seconds literally all round because you have sline vomit gland or you get 200 quadrillion brain damage from incurable brainrot.

ParagonPrime:
I think Abduction SHOULD be modified slightly to make it more interesting and focused-

First, remove all glands that are harmful, or modify them so the benefits stay but the negatives are gone/severely reduced.

Second, add a direct objective to the Abductors- get x many Research Points by the time the round ends. Research Points are gotten simply by abducting and experimenting on people and animals, with different effects:

Monkeys and dead humanoids: 0.5 RP
Living humanoids (Phytosians, Unathi, Preternis): 1 RP
Augmented humanoids, humanoids listed on the crew manifest as Security staff: 2 RP
[...]

PR #721, opened 2016-03-28
Abductor mode

MacHac:
Refer to issue #705.
### Intent of Pull Request

This PR adds two 'new' modes to the game, traitor + abductor and traitor + abductor + changeling.  It also adds them to the secret rotation.  This will allow players to play abductor again, without any of the problems traditionally associated with abductor only rounds.

Feedback encouraged.


Groudonmaster:
I think we should probably fix the Abductor mechanics i.e. the stupid console being broken atm before we add them back in.

X-TheDark:
Indeed, please do not merge this until the abductor mode itself (machines needing 30 seconds to bring icons back, maybe some glands balancing) is fixed/balanced.

Super3222:
traitorabchan

what what

Also I thought you worked out the balance issues with Adam already?

FluffySurvivor:
this means we will see traitors and lings more often (besides abductor)
:-1:

ShadowDeath6:
**DO NOT HAVE THIS IN CONJUNCTION WITH CHANGELING HOLY FUCK**

:-1: x :100: IF THIS IS WITH CHANGELING

MacHac:
@Super3222 The balance issue fixes are not in the code right now, I wanted to get the PR up first.
@ShadowDeath6 Changelings are one of the abductor's weaknesses, and their possible presence could help ensure that the abductors don't get sloppy.  The probe mode on the stun baton can identify them.
@FluffySurvivor The chances could be switched around.

I'm going to see if I can't figure out what the console is doing.

MacHac:
Console is fixed by the latest commit.  I'll look at the balancing factors.

ShadowDeath6:
> probe mode tells who's changeling

Pretty sure it just tells their race, not their antag status.

Also keep in mind that abductor glands can be pure cancer to literally everyone in the round, and there will be a lot of people's antag rounds ruined by this.

MacHac:
No, it does changelings too.  Changelings have a special alert.

MacHac:
All the no-questions-asked round ruining glands are commented out.  Virus is still in because you can fix its effects by curing the one with the organ.

ShadowDeath6:
> curing the one with the organ

I thought you couldn't catch the disease if you had the virus.

X-TheDark:
Ok, seems latest commits removed some pretty annoying glands and fixed the console buttons (and by extension others, since it's in mob/life rather than tied to the console), I think this should be given a go. I'm all for variety and having EVERYONE GET IN HERE sounds like something fun (to at least try).

:+1:

code/game/gamemodes/abduction/traitor_abd.dm
```diff
@@ -0,0 +1,264 @@
+/datum/game_mode/traitor/abduction
+	name = "traitor+abductor"
+	config_tag = "traitorabd"
+	traitors_possible = 3 //hard limit on traitors if scaling is turned off
+	restricted_jobs = list("AI", "Cyborg")
+	required_players = 0    // Should only spawn traitors during extreme lowpop, but wierd things do happen
```
X-TheDark: I'd change this to something like 20 purely to make sure nothing breaks **too** hard.

code/game/gamemodes/abduction/traitor_abd_ling.dm
```diff
@@ -0,0 +1,79 @@
+/datum/game_mode/traitor/abduction/changeling
+	name = "traitor+abductor+changeling" //MADNESS
+	config_tag = "traitorabdchan"
+	traitors_possible = 2 //hard limit on traitors if scaling is turned off
+	restricted_jobs = list("AI", "Cyborg")
+	required_players = 0
```
X-TheDark: I'd change this to 20-30, again, to make sure there's some players to go around.

MacHac:
Yeah, at the very least implement the buttons change.  This should stop shadowlings from having this problem too if they hivemind too often.

@ShadowDeath6 They're given the disease, but with the carrier flag on, so they don't manifest any symptoms themselves.  They can still be cured.

code/game/gamemodes/abduction/traitor_abd.dm
MacHac: Left that in for testing, will change it

code/game/gamemodes/abduction/traitor_abd_ling.dm
MacHac: I would also like to add that abductor teams can't spawn if there aren't at least 6 people without a special role per team, which would regulate things a bit

ShadowDeath6:
How many more people does traitor + abductor need than regular traitor or traitorling?

MacHac:
Regular traitor and traitorling both don't need any actual non-antags, because they can (theoretically) be greentext'd even if you're the only person on the server.   This is why you will always be either traitor or traitorling if you join an empty server.

MacHac:
Are there any other problems?  Any other reasons why this shouldn't be merged?

PR #727, opened 2016-03-29
Adds admin tools for dealing with grief and stuffs.

AsV9:


AsV9:
I will be redoing this with additional buttons.

PR #749, opened 2016-03-31
Deployable Barrier Buff v2

Super3222:

Buffs the deployable barrier.
Version 1: #634 

*Deployable barrier now accepts encryption keys that associate with the security channel (ex: cap, hos, etc.)
*Deployable barriers when taking damage (and when they have an encryption key installed) will report their state to the security channel
*You can attach signalers to the deployable barrier, so they report their locations with their state
*You can now tell how much damage a barrier has taken by examining it.
*Added a signaler to the map, spawns in the security office.
*There's now a click delay when you are spam-clicking the barrier. So they don't blow up so easily now
*Upon detonating, a deployable barrier will report that it's exploding.


code/game/machinery/deployable.dm
```diff
@@ -181,17 +191,163 @@ for reference:
...
+			visible_message("[src] swallows the encryption key and vibrates happily!")
+			attachedradio.set_frequency(SEC_FREQ)
+			radio_freq = SEC_FREQ
+			if(istype(W, /obj/item/device/encryptionkey/headset_sec))
+				installedkey = new /obj/item/device/encryptionkey/headset_sec
+			if (istype(W, /obj/item/device/encryptionkey/heads/captain))
+				installedkey = new /obj/item/device/encryptionkey/heads/captain
+			if (istype(W, /obj/item/device/encryptionkey/heads/hos))
+				installedkey = new /obj/item/device/encryptionkey/heads/hos
+			return qdel(W)
```
Cruix: Why are you making a new key and deleting the old one? Why not just use the old one?
On another note, why are you returning the result of qdel()?
Also, if the key has the security channel but is not one of these types, it will be deleted, never to be seen again.

code/game/machinery/deployable.dm
```diff
@@ -181,17 +191,163 @@ for reference:
...
 				src.health -= W.force * 0.75
 			if("brute")
 				src.health -= W.force * 0.5
 			else
+
+		// announcing damage taking place
+
+		if (attachedradio.frequency == SEC_FREQ && !src.emagged || attachedradio.frequency == SYND_FREQ) // security / syndicate
+			var/thebarrier = "A deployable barrier"
+			message_admins("IT FLEW.")
```
Cruix: Please make this more descriptive, and maybe even add a jump href.

code/game/machinery/deployable.dm
Super3222: I had to break it down to the exact type of encryption key because I was having trouble spawning the variable type.

And I've looked through the encryptionkey.dm, and these are the only three which contain the security channel under `var/list/channels`.

Cruix: It may work right now, but someone might add a new key later, and they should not have to update deployable barrier code as well.

Why are you deleting and re-creating the keys instead of just unequipping them and putting them in the barrier?

Super3222: Oh wow, I'm such an idiot. I over complicate things. Anyway, I'll be fixing this.

ghost:
Make them also allow you to shoot over them if you're standing right next to them.

PR #767, opened 2016-04-02
Floodlamps

TehFlaminTaco:
Adds Flood Lamps

When assembled using up to 3 Flashlights and a power-cell, will emit a powerful light with a 15 meter radius, although quickly drains smaller cells.

Can be ordered from cargo, and one spawns in Processing, that area where you send nerds to go mining.

Useful for mining, because it removes the need for a flashlight. Useful in power outages, for obvious reasons.

Good Shadowling counter that isn't an 'i win button'. Keep areas sectioned off from them, but a sneaky shadowling can still bum-rush it and turn it off.

Not effected by Veil.


ParagonPrime:
_colonial muhreens PTSD_

PR #771, opened 2016-04-02
New brig map

utahclock:
adds new brig layout


utahclock:
http://puu.sh/o3xcZ/5a774587ae.png

PR #798, opened 2016-04-04
adds space scout knife

utahclock:
adds space scout knife, to z3 and z4 for adventuring!


ShadowDeath6:
Do you have a picture of the icon to show?

Notamaniac:
capture

(I'm still waiting for Adam to come by and go TWENTY FORCE WEAPON IN SPACE OP OP WAAAAAAAARRRRGH)

ShadowDeath6:
20 force is pretty legit since there's only one on the entire server I guess

ghost:
20 force weapon isn't too bad but the sprite looks like a hotdog with a grey dick attatched

but serious 20 damage weapon for space exploring is fair, even underpowered a bit

ShadowDeath6:
let's make it 27 damage so it's literally a retracted esword that does regular esword damage :^)

also it totally looks like a hotdog kek

Notamaniac:
Criticising the sprite, I'd change the perspective so it'd look more interesting, rather than a flat on hotdog look. Also it needs shading, otherwise it'll stick out like a sore thumb. So basically like this but less shitty:
shitknife

AsV9:
/cooonflicts/

don't reuse the changelog files damnit

Issue #777, opened 2016-04-03
Wizard / Borer bug


LesVegetables:
### Expected Behaviour

Lich Wizards respawn without their borer
### Actual Behaviour

If you die as a lich with a borer inside of you, you respawn catatonic and cant get in your body
### Steps to reproduce the problem
1. die with a borer
2. get ghosted
3. your body respawns but you arent in it

other notes: the borer only took control once for like 2 seconds and was not in control when i died

PR #799, opened 2016-04-05
Minor borer changes

ShadowDeath6:

Fixes issue #777 by removing borers being able to enter the skeleton race.

Also changes up some of the span class formats to make borer receiving messages prettier. 

Adds a message to the host being infested so they know when to run away.

Also teaches new borers the secret of *talking while in a host.

Also ALSO: fixes the spelling mistake with "telepathicaly" to "telepathically"

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_powers.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_powers.dm
@@ -4,7 +4,7 @@
 	set desc = "Infect a adjacent human being"
 
 	if(victim)
-		src << "<span class='boldnotice'>You already have a host! leave this one if you want a new one.</span>"
+		src << "<span class='warning'>You already have a host! leave this one if you want a new one.</span>"
 
 	if(stat == DEAD)
 		return
@@ -18,14 +18,15 @@
 	if(!H) return
 
 	if(H.borer)
-		src << "<span class='usernotice'>[victim] is already infected!</span>"
+		src << "<span class='warning'>[victim] is already infected!</span>"
 		return
 
 	if(CanInfect(H))
-		src << "<span class='boldnotice'>You slither up [H] and begin probing at their ear canal...</span>"
+		src << "<span class='warning'>You slither up [H] and begin probing at their ear canal...</span>"
 		src.layer = MOB_LAYER
+		H << "<span class='userdanger'>You feel something trying to enter your ear...</span>"
 		if(!do_after(src,30))
-			src << "<span class='boldnotice'>As [H] moves away, you are dislodged and fall to the ground.</span>"
+			src << "<span class='warning'>As [H] moves away, you are dislodged and fall to the ground.</span>"
 			return
 
 		if(!H || !src) return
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -405,8 +406,8 @@ mob/living/carbon/proc/release_control()
 
 		var/mob/living/simple_animal/borer/newborer = new(get_turf(src))
 		newborer.transfer_personality(C)
-		visible_message("<span class='userdanger'>[src] heaves violently, expelling a rush of vomit and a wriggling, sluglike creature!</span>")
```
oisin100, line 408: I think this one should have been kept userdanger.

DavidEdwards:
> Adds a message to the host being infested so they know when to run away.

Keep it sneaky. Otherwise borers will have no chance.

code/modules/mob/living/simple_animal/borer/borer_powers.dm
ShadowDeath6, line 408: It's shown to everyone around, so it's kind of annoying to see a giant red message which basically gets the baby borer killed and incriminates the person as well. A smaller message means a tiny bit more stealth which is in my opinion a good thing.

ShadowDeath6:
They have paralyze host for a reason.

Time-Green:
Shadow, they removed that.

ShadowDeath6:
No they didn't.

oisin100:
> They have paralyze host for a reason.

Paralyze shows a message ruining their stealthiness. Borers should be able to sneak into their host as long as they dont use paralyze. Keep infecting silent.

ShadowDeath6:
The point of that is to keep players from bitchmoaning about a borer getting into their head while showing literally no indication other than it appearing for about a second and a half before teleporting into their head. It's only fair to give players a heads-up that a borer will infect them. 
Besides, if you time it right, paralyze will give enough time to infect them.

code/modules/mob/living/simple_animal/borer/borer_powers.dm
oisin100, line 408: > Adds a message to the host being infested so they know when to run away.

Ye, Real stealthy.

If you want your borer to survive reproduction go to some secluded place and do it. Borers are a "threat" to anyone nearby them thus the userdanger message. Its already easy pez to reproduce, dont make it any easier.

AsV9:
Unless @ShadowDeath6 can cough up some support for this, it will be closed in 24hrs.
I'm personally opposed to most of these changes.

ShadowDeath6:
What are you opposed to, asv? I can remove the message from infesting but the span classes for old borer things were really ugly and wizard still gets bugged  in the very unlikely case it makes a borer.

The message encourages borers to actually play smart and try to stun a host instead of relying on complete invisibility in text logs and then making someone salty because there's literally no warning other than the HUD that pops up. Bay did it, it worked fine.

AsV9:
We're not bay. I'd like borers to be stealthy and be able to infiltrate hosts without a giant userdanger message popping up encouraging them to bee-line to the chemist to make sugar.

ShadowDeath6:
It's not like they don't get an HUD over their head and a borer talking to them. It would give people a chance to actually run away when a borer tried to infect them, which is only fair. I can make it smaller if you think it'll be that much of a problem.

DavidEdwards:
I'm honestly not sure we should have a HUD added for the borer host either, unless the borer allows it.

That is something that has bugged me a little bit.

AsV9:
I'm kinda with Kn0ss0s on that issue aswell. The borer should have the HUD, not the host, forcing them to communicate if they want to share that information.

ShadowDeath6:
I think that was originally indended. Does that not happen?

ShadowDeath6:
tl;dr people want borer messages when I ask but nobody stands up for me when I do if

I'll comment it out in case somebody wants it later.

ShadowDeath6:
I need to sleep, I hope I've appeased you. ;_;

Issue #801, opened 2016-04-05
Ian is a 4th dimensional being


ToGWtF:
No but really. Ian can go through airlocks and blast doors. Even windoors, but only in one direction so he cant go back out of them
# Feature?

ShadowDeath6:
Ian was intended to have the same pass flags as drones, so yes, this was completely intended

Groudonmaster:
It's still an issue then because windoors only allow Ian/Drones to go through one way then they're stuck

ghost:
feature

DavidEdwards:
Dont mess around Adam, this is a genuine bug if someone gets stuck because of it.

ShadowDeath6:
Drones don't have that problem since all access and Ian can't get on tables by himself, so in the very unlikely case that he gets tabled and voluntarily moves through the windoor, he won't be able to get back, but this is probably extremely uncommon.

ShadowDeath6:
@X-TheDark

Groudonmaster:
Yeah but wasn't the point of adding the drone flags was to stop doors from opening?

X-TheDark:
Yup, problem with windoors only checking these flags 1-way, so this actually is a bug that is specific with how windoors check for pass_flags defined on mobs. You can go in, but not go out without opening the windoor.

PR #821, opened 2016-04-07
Drone passing through door fixes (+drone toolbox tweaks)

### LINKED ISSUE #817: Picking up drones
### Expected Behaviour

When a player _attempts_ to pick up a drone they should pick it up after a delay unless the drone moves out of range.
### Actual Behaviour

Once a player attempts to pick up a drone no matter how far away the drone runs as long as the player stays still they'll succeed in picking up the drone
1. Attempt to pick up a drone.
2. Have the drone move out of range before you finish picking them up.
3. Watch as the drone appears in your hand anyway.

X-TheDark:
Refer to issue #801 
Also fixes #817 (might as well, as I'm messing with drones)

Fixes some doors not passing drones through (such as one-way window door thing). Also allows drones to pass through Firedoors (the ones that drop during fire) as well as Plastic Flaps.
- Drone toolbox have 10 slots, it is no longer dropable, though. Max combine weight class changed accordingly.
- Adds a pocket fire extinguisher to the toolbox (meaning 2 slots free).
- Renamed the toolbox to something more telling, now that it's not dropable.
- EDIT: Also fixes the remote drone pick-up bug.

### DIFF
--- a/code/game/machinery/doors/windowdoor.dm
+++ b/code/game/machinery/doors/windowdoor.dm
@@ -74,7 +74,7 @@
 	return
 
 /obj/machinery/door/window/CanPass(atom/movable/mover, turf/target, height=0)
-	if(istype(mover) && (mover.checkpass(PASSGLASS) || mover.checkpass(PASSDOOR)))
+	if(istype(mover) && (mover.checkpass(PASSDOOR) || mover.checkpass(PASSGLASS)))
 		return 1
 	if(get_dir(loc, target) == dir) //Make sure looking at appropriate border
 		return !density
@@ -92,7 +92,7 @@
 	return !density || (dir != to_dir) || check_access(ID)
 
 /obj/machinery/door/window/CheckExit(atom/movable/mover as mob|obj, turf/target)
-	if(istype(mover) && mover.checkpass(PASSGLASS))
+	if(istype(mover) && (mover.checkpass(PASSDOOR) || mover.checkpass(PASSGLASS)))
 		return 1
 	if(get_dir(loc, target) == dir)
 		return !density
--- a/code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
+++ b/code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
@@ -81,7 +81,7 @@
 			user << "<span class='warning'>[src] is buckled to the [buckled.name] and cannot be picked up!</span>"
 			return
 		user << "You start picking up the [src]."
-		if(!do_after(user, 20, target = user))
+		if(!do_after(user, 20, target = src))
 			user << "You failed picking up [src]."
 			return
 		user << "<span class='notice'>You pick [src] up.</span>"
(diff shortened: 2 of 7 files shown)


Super3222:
Finally, drones are now finally DRONES. 

@X-TheDark, one problem. When the drone dies, it will drop the toolbox. The next unlucky person will pick it up will be unable to drop it, I believe.

X-TheDark:
Let me check that real quick. Also, drones cannot move the toolbox from their internal storage slot to their hand.

X-TheDark:
Nope, killing the drone doesn't drop it's toolbox, because it's in the internal_storage slot and you cannot move the toolbox out of it. Just tested this in-game by killing the bot.

code/game/machinery/doors/windowdoor.dm
```diff
@@ -74,7 +74,7 @@
 	return
 
 /obj/machinery/door/window/CanPass(atom/movable/mover, turf/target, height=0)
-	if(istype(mover) && (mover.checkpass(PASSGLASS) || mover.checkpass(PASSDOOR)))
+	if(istype(mover) && (mover.checkpass(PASSDOOR) || mover.checkpass(PASSGLASS)))
```
Cruix, line 77: Swapping the order of these should not make a difference.

code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
```diff
@@ -81,7 +81,7 @@
 			user << "<span class='warning'>[src] is buckled to the [buckled.name] and cannot be picked up!</span>"
 			return
 		user << "You start picking up the [src]."
-		if(!do_after(user, 20, target = user))
+		if(!do_after(user, 20, target = src))
```
Cruix, line 84: :+1: "Target = user" made me chuckle.

code/game/machinery/doors/windowdoor.dm
X-TheDark, line 77: Correct! For some reason my OCD in this particular case demanded that I started with the stronger clause (PASSDOOR), since it's primarily a door, rather than PASSGLASS check, since the fact that it's made of glass is secondary. Call it a personal quirk that spurred up in the moment of coding.

code/modules/mob/living/simple_animal/friendly/drone/interaction.dm
X-TheDark, line 84: Honestly, perhaps this should just be rewritten as !do_after(src, 20, target = user), since really...target only denotes ON WHOM the do_after process bar thing is displayed, so perhaps it'd make sense for the guy who tries to pick someone up to see the progress bar on whatever they are trying to pick up (the drone, as was the case with target = user) rather than themselves (as is the case with target = src). Obviously, it's pretty much irrelevant and is a case of apple and orange(M,3), but still.

ParagonPrime:
No more dropping toolbox? But how will post-round drone griff work now? :(

PR #825, opened 2016-04-08
Adds new blob type, Disorienting Fluid.

ShadowDeath6:

Causes sensory disruption, such as blindness for a very short amount of time and blurry vision similar to pepper spray or spore toxin's effect, and deafness. It also does very low toxin and radiation damage. It /technically/ does brute because every blob attack works that way. 
Spores also inflict the status effects. 

You need to get a lot of hits off for the rad damage to actually take effect, so I expect this blob to be more of a "long haul" kind. 

Thanks @Groxic for the help with blob code once again.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Blob-Reagents.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Blob-Reagents.dm
@@ -619,11 +619,13 @@
 
 /datum/reagent/blob/disorienting_fluid/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message, touch_protection, mob/camera/blob/O)
 	reac_volume = ..()
-	M.irradiate(0.2*reac_volume)
-	M.apply_damage(0.2*reac_volume, TOX)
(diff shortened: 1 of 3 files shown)


ShadowDeath6:
uh oh

ParagonPrime:
Seems a bit underpowered. 

Remember that we're "low arrghpee". That deaf-blind assistant doesn't stop whacking you with a welder just because he can't see for a bit.

ShadowDeath6:
> attacking blob after being hit multiple times
> get fucked by radiation
> get permastunned from stamina by radiation
> ayy

ShadowDeath6:
In hindsight, 10 toxins per hit is pretty low, I'll up it to 15 when I get time.

Cruix:
Keep the toxin damage as it is and give it a chance to give the clumsy mutation. Laugh as security shoots themselves in the foot over and over.

ShadowDeath6:
ready for merge

ghost:
MEEEERRRGEEE

PR #830, opened 2016-04-08
adds new brig

utahclock:
adds new brig layout


Groudonmaster:
What's with the random rwalls by interrogation?

utahclock:
those can be  changed to normal walls

EXOTICISME:
So many cells. Can you put firing range instead? With practice laser and such. Also the empty space at labour exit can be decorated with plants or chair for whoever want to watch the prisoner suffer.

EMT321:
Does the detectives office get a shutter? Otherwise people can just break the glass and run in.

utahclock:
walls change, fixed door where it was not supposed to be, added shutters to det office.

ghost:
why is det office in sec

det is not a sec officer he's a PI get his office out of sec it's not supposed to be there for a reason

same with the lawyers

utahclock:
@AdamElTablawy they are in brig not sec office. the lawyer is where its at so he can over see inmates. det is where its at because he needs to be able to investigate, evidence and stuff.

utahclock:
@AdamElTablawy det and lawyer have same access as they do know which is innerbrig.

FluffySurvivor:
Sec equipment room accesible by maint, not even protected by rwalls. 

nuff' said

Groudonmaster:
Don't worry Utah I think you're doing a fine job

Issue #528, opened 2016-03-11
ADAM HAS ANOTHER GIANT FUCKLOAD OF IDEAS-ENGINEERING EDITION


ghost:
So, there are a lot of issues wrong with engineering.

Specifically, engineering does not have much to do.

They set up the power, go braindead for 40 minutes, and then a bomb blows up the bridge and it's so tiring to fix nobody does it and just calls the shuttle.

So, I have devised multiple different systems relating to engineering to keep them involved in the round and no longer so insular. Divided into 3 different sections-_Department reworks_, _Management_, and _Events._

---
## Department reworks.

R&D has, to be frank, too much to do. It's ridiculous that science is responsible for upgrading and building machines that engineering is supposed to do. 

R&D is supposed to be developing new tools, yes, but not responsible for EVERYTHING. Engineering is the UPGRADES team, R&D is the DEVELOPMENT team.

So, to promote department interconnectivity-
### 1. Remove the 'stock parts' ability from R&D. R&D can no longer upgrade machines on it's own with RPED's.

Instead, add a 'parts manufacturer' to engineering. As science's levels go up and they sync, engineering can sync servers with R&D, thus granting them access to the nano-pico-femto tiers of parts. 

Engineers are now responsible for printing and upgrading machinery, and R&D is responsible for doing the work to grant them access to these parts. Engineers have more to do, and more department interconnectivity is required.

'BOO HOO I DON'T WANT TO WAIT 5 EXTRA MINUTES WAAAAAAH'

R&D isn't required anyway. It is a frivolity. At least now, engineering has more to do except resolve inane antagonist actions, and guess what, if you can't vomit out infinite resources 5 minutes into the round, now you know how everyone else feels when you turtle up in science behind 7 automatic dildo printers.
### 2. Remove the circuit imprinter from R&D and add it to engineering.

 Engineering is responsible for the construction of machinery and gear, science is responsible for devising it. Now, R&D has to actually leave science to get research up, meaning you can't powergame-beeline it anymore, and that R&D can't turtle up with their own private autolathe and exosuit fabricator and cloner on their own anymore. 

[...]

PR #840, opened 2016-04-09
Added partlathe to code. Adminspawn only.

Cruix:
Small part of issue #528

Added the partlathe to the code. It does not exist in-game and cannot be created except by adminspawn. The partlathe can sync to the research network and acts like an autolathe, but can only build stock parts and metal/glass sheets. It cannot currently be hacked.


ghost:
Why do we need more lag?

ghost:
We're building to 528 and it causes no lag.

PR #863, opened 2016-04-11
Adds random event for frying APCs.

AsV9:
I really wanted to add increased chances for AI core and AI upload..

I somehow resisted my temptations.
### How this shit works:

Random event fires.
This can happen pretty early in the round. Like almost at roundstart.

A random roll is done in the range 2 to 5. This decides the percentage of all APCs on the station gets fried.

Somewhere between 15 and 25 seconds after that, the actual thing happens.

5 seconds after that, a random roll is done to find out if we should alert the crew. This is currently set to 30% of the time.

If we alert the crew, a big message telling everyone appears.
If not, the admins are alerted, so they don't get confused.

### DIFF
--- a/code/modules/events/apc_fry.dm
+++ b/code/modules/events/apc_fry.dm
@@ -1,0 +1,24 @@
+/datum/round_event_control/apc_fry
+	name = "Fried APCs"
+	typepath = /datum/round_event/apc_fry
+	weight = 70
+	earliest_start = 1000
+
+/datum/round_event/apc_fry
+	announceWhen = 15
+	var/fry_chance = 3
+
+/datum/round_event/apc_fry/setup()
+	startWhen = rand(15, 25)
+	fry_chance = rand(2,5) //2 to 5 percent of station APCs get fuck'd.
+	announceWhen = startWhen + 5
+
+/datum/round_event/apc_fry/announce()
+	if(prob(30))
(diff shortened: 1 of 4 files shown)


ShadowDeath6:
Frying meaning they lose all their charge or just turn blue? In the elec-storm event an APC can fry which turns it blue and permanently sets its charge to 0 until it's fixed. If that can happen in the AI chamber...

AsV9:
It does the same as swiping an emag does. Changes icon to the blue blinking, unlocks it.

Super3222:
So who do I blame when I'm in the middle of deconstructing something important and the APC gets fried right away.

AsV9:
The APC isn't unusable, it's just perma-unlocked and blinks blue. You can always deconstruct it and make it work again.

code/modules/events/apc_fry.dm
```diff
@@ -0,0 +1,24 @@
+/datum/round_event_control/apc_fry
+	name = "Fried APCs"
+	typepath = /datum/round_event/apc_fry
+	weight = 70
```
oisin100: 70 weight seems a bit extreme doesn't it? Electrical storm is only 40 and thats frustrating enough.

AsV9: Huh, you're probably right. I looked at some other events which I never see firing and they had like 70-80 weight. Scaling it down.

PR #872, opened 2016-04-12
Fixes nuke disk pinpointer bug.

### LINKED ISSUE #813: Pinpointer doesn't show right position when disk is in the nuke.
### Expected Behaviour

If you space the nuke with the disk inside, the pinpointer should show where the new nuke disk spawns.
### Actual Behaviour

It shows a question mark even after turning it off and on after the new disk is spawned. This only happens when the old disk is spaced inside of the nuke.
1. Put the disk in the nuke
2. Space it
3. Watch as your pinpointer gets broken for eternity.

AsV9:
Fixes #813 

This is the 2nd time I've fixed this bug. The map portion of the fix was reverted by Hardy who refused to properly merge the map.


AsV9:
oops forgot to rename pull.

PR #873, opened 2016-04-12
Security Barriers will report how much health they have left

Super3222:

Simple tweak. Security barriers will report how much health they have when taking damage. This will only work if you have secured an encryption key into them.

Also fixes a small, but possible bug. Barriers will no longer report damage when something that does 0 damage hits them.

This will not make security "OP". Remember these are GIANT barriers meant to keep things/places safe and completely secure.

Also emags still currently disable radio signals from leaving a barrier, so that's a pretty good counter.

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -326,20 +326,22 @@ for reference:
 
 			var/damagereport
 			if(W.force <= 5 && W.force != 0) // less or equal to 5
-				damagereport = "a small amount of damage from an external source"
+				damagereport = "a small amount of damage."
 			if(W.force > 5 && W.force <= 15) // in the 5-15 range, but greater than 5
-				damagereport = "a moderate amount of damage from an external source"
+				damagereport = "a moderate amount of damage."
 			if(W.force > 15) //greater than 15
-				damagereport = "a large amount of damage from an external source"
+				damagereport = "a large amount of damage."
+			else if (!damagereport)
+				return
 
 			if(src.health <= 95 && src.health > 50)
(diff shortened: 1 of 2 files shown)


code/game/machinery/deployable.dm
```diff
@@ -326,20 +326,22 @@ for reference:
...
-				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] in a sufficient state!!",radio_freq)
+				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] The barriers rate of functionality is at [health]%!!",radio_freq)
 
 			if(src.health <= 50 && src.health > 25)
-				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] while they are in need of dire repair!!",radio_freq)
+				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] The barriers rate of functionality is at [health]%!!",radio_freq)
 
 			if(src.health <= 25 && src.health > 0)
-				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] at a currently fatal state!!",radio_freq)
+				attachedradio.talk_into(src, "Alert! [thebarrier] is suffering [damagereport] The barriers rate of functionality is at [health]%!!",radio_freq)
```
ShadowDeath6, line 344: are you able to use apostrophes, or no? it doesn't look very good withot them.

Super3222, line 344: @ShadowDeath6 
I am. However read the the damage report variable, and how it is assigned.

PR #893, opened 2016-04-14
Fixes a few missing cameras from the map.

AsV9:
Two cameras at engineering fixed. Quite minor.


00aquir:
Um, right now the north east engine camera is called "North west".

AsV9:
That, I didn't think of.

Issue #664, opened 2016-03-23
Revenants up and die at random when using abilities


nazitra:
### Expected Behaviour

Be spoopsy ghost, fly around shocking lights and making clonepod soup.
### Actual Behaviour

Die even if you have more than enough essence to use your abilities. No clonepod soup for you.
### Steps to reproduce the problem
1. Be ghost
2. Pray to RNG, become revenant
3. Die, rage to the heavens

Time-Green:
I fully agree. I had enough essence and died the same way. And then Donny took it....

nazitra:
...only to die doing the same thing, trying to shock lights. >:(

ShadowDeath6:
> not using nothing but transmit to fuck with people

Wardog390:
please fix.

Time-Green:
Ok, I had 100 essence and bought defile for 30e
And then died. Like whhhyyyy?????

Greatbigben:
is a fix in progress?

penguinfisher:
can confirm that defile is fuked. I had 45E and bought defile for 35 after having all my other abilities and just up and died.

X-TheDark:
This might be related to some lag issue, as it's impossible to reproduce on a local server (tried for 1 hour). So currently, I'm not sure if this can be fixed, as the reason for this cannot be isolated. Will try to find something.

X-TheDark:
Unfortunately, this seems to be just plain impossible to reproduce locally. Please reopen the issue **if you find any additional information under what circumstance beyond "at random when using/buying abilities" this happens.**

PR #902, opened 2016-04-15
Small fixes to revenant from /tg/

X-TheDark:
Refer to issue #664 
All it takes is for a coder's fun to be ruined for that said coder to become motivated to fix the issue as soon as possible...go figure.

Minor port of some small stuff from /tg/, credit to original creators.
Refer to tgstation/-tg-station/pull/11757 by ghost - death() proc no longer triggers on non-revealed revenants.
Not linking because the actual pull from /tg/ touches upon many, many more things.
1. ..() call moved to end of Life()
2. Ported over checks for revenant being revealed or dead for death().
3. Also uses MSO's stoplag() instead of sleep().
4. Definately fixes revenants dieing randomly. **Cause** : Revenants were not healed correctly when not revealed.
5. Revenants can now quick-cast Transmit via Control Click.
6. Turrets no longer target non-revealed revenants.
7. Also sprucens up the turret code by adding calculate_targets() from /tg/

Stuff tested:
- Revenant can still be killed while it is revealed.
- Revenant doesn't die if it buys something after it's been damaged and becomes concealed.
- Quick-casting Transmit.
- Revenants now properly heal brute/fire damage when not revealed.

### DIFF
--- a/code/modules/mob/living/simple_animal/revenant/revenant.dm
+++ b/code/modules/mob/living/simple_animal/revenant/revenant.dm
@@ -80,13 +80,7 @@
 		AltClickOn(A)
 		return
 	if(modifiers["ctrl"])
-		if(istype(A, /mob/living))
-			var/mob/living/LM = A
-			var/list/targets = list()
-			targets += LM
-			revtransmit.cast(targets)
-		else
-			CtrlClickOn(A)
+		CtrlClickOn(A)
 		return
 
 	if(world.time <= next_move)
@@ -95,6 +89,13 @@
 	if(ishuman(A) && in_range(src, A))
 		Harvest(A)
 
+/mob/living/simple_animal/revenant/CtrlClickOn(atom/A, params) //Copypaste from ghost code - revenants can't interact with the world directly.
+	if(istype(A, /mob/living))
+		var/mob/living/LM = A
+		var/list/targets = list()
+		targets += LM
+		revtransmit.cast(targets)
+	..()
 
 /mob/living/simple_animal/revenant/proc/Harvest(mob/living/carbon/human/target)
 	if(!castcheck(0))
--- a/code/modules/mob/living/simple_animal/revenant/revenant.dm
+++ b/code/modules/mob/living/simple_animal/revenant/revenant.dm
@@ -44,7 +44,6 @@
 	var/list/drained_mobs = list() //Cannot harvest the same mob twice
 
 /mob/living/simple_animal/revenant/Life()
-	..()
 	if(essence < essence_min)
 		essence = essence_min
 	if(essence_regenerating && !inhibited && essence < essence_regen_cap) //While inhibited, essence will not regenerate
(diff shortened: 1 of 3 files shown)


ghost:
Does this fix revenants randomly dying when using abilities

X-TheDark:
Hopefully, yes, as the "revenant dieing randomly" is a general problem. Calculating/refreshing health before calling any other Life() proc up the stack should fix this, but there's no way to be certain, as I am unable to test this properly locally.

Kmc2000:
Does this fix revs being targeted by ai turrets when hidden? that's a bitch.

X-TheDark:
Oh, damn, that's a rather good point, actually. Let me see if I can add it in.

X-TheDark:
Ok, this is ready to be merged now.
**Cause of random death** : Revenant was not properly healed while being concealed. This has been fixed.

**Features added by last commit**
- Revenants can now quick-cast Transmit by Control+Clicking. No more chasing that meth'ed up chemist to talk to him.

code/modules/mob/living/simple_animal/revenant/revenant.dm
```diff
@@ -77,7 +80,13 @@
 		AltClickOn(A)
 		return
 	if(modifiers["ctrl"])
-		CtrlClickOn(A)
+		if(istype(A, /mob/living))
+			var/mob/living/LM = A
+			var/list/targets = list()
+			targets += LM
+			revtransmit.cast(targets)
```
Cruix: Put all this in CtrlClickOn() instead of ClickOn(), that's the only reason CtrlClickOn() exists.

Apart from that it's good shit :+1:.

ghost:
revenants are INTENDED to be shot by turrets when invisible

code/modules/mob/living/simple_animal/revenant/revenant.dm
```diff
@@ -95,6 +89,13 @@
 	if(ishuman(A) && in_range(src, A))
 		Harvest(A)
 
+/mob/living/simple_animal/revenant/CtrlClickOn(atom/A, params) //Copypaste from ghost code - revenants can't interact with the world directly.
+	if(istype(A, /mob/living))
+		var/mob/living/LM = A
+		var/list/targets = list()
+		targets += LM
+		revtransmit.cast(targets)
```
X-TheDark: Should I return after this line?

code/modules/mob/living/simple_animal/revenant/revenant.dm
```diff
@@ -95,6 +89,13 @@
...
+	..()
```
X-TheDark: Should this be removed? Do I need/not-need to call the parent proc, I wonder.

code/modules/mob/living/simple_animal/revenant/revenant.dm
X-TheDark: Thanks, will do.

X-TheDark:
Nah, I think this is good.

Issue #894, opened 2016-04-14
Eswords need armor pen


FluffySurvivor:
### Expected Behaviour

Eswords being at least a bit decent against multiple armored hostiles.
### Actual Behaviour

They are trash.
### Steps to reproduce the problem
1. Call ERT as tator because fluffy is nice
2. 1v5
3. Hit the fuckers 30 times, not even one of them drops.

X-TheDark:
> 1v5

I think I've found the problem and unfortunately, this probably shouldn't be fixed by code. Making E-Sword capable of beating odds like that is something that should never happen.

FluffySurvivor:
yeah but it takes an eternity to even slash a sec officer to death

the cost efficiency of eswords dramatically drops when used against armored targets

X-TheDark:
Also, for reference:
- Attacking naked human = 30 damage.
- Attacking ERT armor person thing = 21 damage

ghost:
armor is literally fucking paper

but sure, give eswords some level of AP

ShadowDeath6:
Le beam of light penetrating opaque materials

Just aim for arms or legs on sec, it might not do 30 damage but it'll kill them a lot faster than aiming for chest. 

If this becomes a 6 fucking tc item it should not be MORE powerful than it already is.

ParagonPrime:
Lightsabers are lightsabers and lightsabers cut through shit ez.

They should be able to have decent armour penetration. And, of course, be much, much harder to disarm when activated.

FluffySurvivor:
1. Lightsabers (and eswords ofc) actually emit plasma. Hot af plasma. Not light. 
2. I used it (a double esword actually) in combo with tasers, the situation where it was supposed to excel, but it let me down.

doerakpoes:
Dubble E swords can defeat 1v5 atleast the odds are better

ParagonPrime:
woah fluffy tone down the starwars fanboy you're making me look inadequate

_ahem_

either way it should be able to cut through fucking... What the fuck is our "standard" armour made of anyway?

X-TheDark:
As Adam said - Paper. That's the standard (t)issue material.

Super3222:
> That's the standard (t)issue material.

that joke is what the issue is asking for... punnetrating armor

ToGWtF:
[...]

PR #912, opened 2016-04-16
Gives eswords a bit of armor pen

FluffySurvivor:
#894

now the plasma sticks can slightly say fuck you to _paper_ armor

this is how it works:

> sec jumpsuit is 10 brute resist. giving eswords armour_penetration = 10 negates that and makes it do the full 30 damage rather than ~20-ish.
> 
> Means, if you target arms or legs, you do more damage than if you target chest because the armor vest is an extra 25 brute resist


ShadowDeath6:
Why are we buffing a 6 tc item to kill sec in 4 hits again?

FluffySurvivor:
because it's shit at times where it's supposed to excel ?

a lower cost won't make it more powerful.

And as always, if anything turns out to be broken it will be nerfed.

Did I mention you don't need a passive "hard to kill" buff ? If you're skilled, you don't need armor.

Cruix:
Because sec has tasers and stunbatons and in a given round at least 10% of sec is competent enough to use them.

code/game/objects/items/weapons/melee/energy.dm
```diff
@@ -94,6 +95,7 @@
 	if (active)
 		force = force_on
 		throwforce = throwforce_on
+		armour_penetration = armourpenetration_on
```
X-TheDark: No need to define another variable.
```
if(active)
    armour_penetration = 10
else
    armour_penetration = 0
```

code/game/objects/items/weapons/melee/energy.dm
```diff
@@ -1,6 +1,7 @@
 /obj/item/weapon/melee/energy
 	var/active = 0
 	var/force_on = 30 //force when active
+	var/armourpenetration_on = 10 //so you can't deal high damage when deactivated
```
X-TheDark: And yet, you never set it back to 0 when it's deactivated.

code/game/objects/items/weapons/melee/energy.dm
Cruix, line 99: You also need to set the armor_penetration to 0 when the sword turns off, otherwise toggling it once will give it that forever. I would recommend having an armorpenetration_off variable just for consistency, even though it will pretty much always be 0.

FluffySurvivor, line 99: I think it should work now.

Cruix, line 99: :+1:

ghost:
Unless sec is actually autistic drawing an esword on an officer gets you harmbatonned and spaced if not stunned, and if they are stunned they're dead regardless of armor, so heah, not changing much here

FluffySurvivor:
I just tested it, 10 armor pen = 3 more damage vs any kind of armor. Nothing huge.

PR #914, opened 2016-04-16
adds several new ores

utahclock:
adds coal,adamantine,clarenkite,platnium,electrite,copper


TheSadFox:
You've made them be generated on the asteroid, even tho they have no use but coin making and tech levels? Why? Why do we need more ores in first place?

> Also 7393 line changes

utahclock:
its part of mining rework . other stuff will be added later

TheSadFox:
Ah, just saw that its a PR for yogstation13:mining-rework.

PR #919, opened 2016-04-17
Make Rev Tolerable Again

MacHac:
Refer to all of the many issues regarding rev.
### Intent of Pull Request

This pull request tweaks how revolutionaries convert people to their cause.
#### CONVERSION TOOLS

Revolutionaries are no longer able to just pick up a flash, flip the switch on the side to 'CONVERT TO NT HATING COMMUNIST', and start zapping away.  Instead, this conversion is done with the use of a conversion tool, or 'Suspicious Device'.  The suspicious device behaves much like a regular flash; However, it looks unique both dropped and in hand.  It also needs time to recharge after converting someone to communism, and will be unusable during this time.
###### But how do revs get more conversion tools?

Head revs now also start with a tool designed to hijack NT flashes and turn them into conversion tools.  This device, when used on a regular flash, will convert it over a period of about thirty seconds.  Note that this process is aborted if the revolutionary moves during this time.
#### BUT WHY
- Revheads can no longer run into a crowded room and have everyone flashed before anyone can react.
- Revheads now have to be at least a little subtle when converting, at least until the revolution gains speed.
- You can now very easily tell the difference between a revhead and someone who has a flash for self defense purposes.
- The fact that most revheads will be carrying around flash hijackers makes it more feasible for staff to find them.

### DIFF
--- a/code/game/gamemodes/revolution/rev_flash.dm
+++ b/code/game/gamemodes/revolution/rev_flash.dm
@@ -1,0 +1,81 @@
+#define REVPEN_COOLDOWN 300
+
+/obj/item/device/flash/rev
+	name = "suspicious device"
+	desc = "You aren't sure what this thing does, but you're almost certain that it can't be good."
+	icon_state = "memorizer"
+	item_state = "nullrod"
+
+	var/cooldown = 0
+
+/obj/item/device/flash/rev/attack(mob/living/M, mob/user)
+	if(!try_use_flash(user))
+		return 0
+
+	if(cooldown)
+		return 0
+
+	if(iscarbon(M))
+		flash_carbon(M, user, 5, 1)
+		return 1
+
+	else if(issilicon(M))
+		add_logs(user, M, "flashed", src)
+		M.flash_eyes(affect_silicon = 1)
+		M.Weaken(rand(5,10))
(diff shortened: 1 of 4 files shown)


ToGWtF:
speedmerge

FluffySurvivor:
DONT KEEP IT METASHIELDED. 

Change the name to "Syndicate Brainwashing Tool". 

This is a major problem with current rev, and that is the lack of ability to act from the loyalists. You won't fix anything if you just make rev rounds longer with the same outcome and gameplay. 

After all, why would the rev heads be careful when they can just press F1 and complain if someone saw them using the flash?

ParagonPrime:
wasn't really at ALL what was suggested on the issue on how to fix rev, but aight, let's see how this works out anyway

in any case, this fixes only the snowballing part- not the murderboning and lack of strategy beyond "bumrush brig, convert cargo, get all-access from HoP, murdermurdermurder"

Super3222:
This is interesting. Let's see how rev goes when they have to slowly convert the crew altogether.

I also agree with what @FluffySurvivor said, and possibly when you are hijacking a flash make the description something suspicious?

I'd like to suggest making the flash hijacker, suspicious device, and the rev flash cost some decent amount of illegal tech. (For revealing that they are syndicate tools)

X-TheDark:
As a memer would say:
:+1: x :100: 

Great job!

PR #931, opened 2016-04-17
Adds client verb to view antag weight

MacHac:
### Intent of Pull Request

Adds a verb which displays information about your current antagonist weight.  This includes your exact weight, and an approximation of how many rounds you'll have to wait to get antagonist.

I'm an SQL noob, so someone needs to check my work here.  I can't really test it because I don't have DB access.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -350,6 +350,10 @@
 				var/DBQuery/query = dbcon.NewQuery("UPDATE [format_table_name("player")] SET `antag_weight` = [weight] WHERE `ckey` = '[ckey]'")
 				query.Execute()
 
+		for(var/client/C in clients)
+			C.last_cached_weight = output[C.ckey]
+			C.last_cached_total_weight = total
+
 		// If after all this we still don't have a candidate, then use the legacy system
 		if(!final_candidate)
 			return pick(antag_candidates)
--- a/code/game/verbs/weightinfo.dm
+++ b/code/game/verbs/weightinfo.dm
@@ -12,22 +12,32 @@
 
 		var/DBQuery/q = dbcon.NewQuery("SELECT `antag_weight` FROM [format_table_name("player")] WHERE `ckey`='[key]'")
 		var/DBQuery/q_total = dbcon.NewQuery("SELECT `antag_weight` FROM [format_table_name("player")] WHERE `ckey` IN ('[ckey_for_sql]')")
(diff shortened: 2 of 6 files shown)


ghost:
M E R G E
E
R
G
E

FluffySurvivor:
HOLY SHIT

:+1:  :+1:  :+1:

S P E E D M E R G E

Kmc2000:
:yes: 

Gimmegimmegimme

TehFlaminTaco:
Many strings aren't sanitized, EZ as fuck SQL injection when?

oisin100:
> Many strings aren't sanitized, EZ as fuck SQL injection when?

Cant use special characters in Ckeys.

Cruix:
sanitize it anyway, it's a big enough danger. If someone can find out how do it they could literally drop every single table in the database, or ban everyone, or set random values to "faggot".

AsV9:
This has to be sanitized before I merge this.

oisin100:
> sanitize it anyway, it's a big enough danger. If someone can find out how do it they could literally drop every single table in the database, or ban everyone, or set random values to "faggot".

None of the SQL statements are sanitized. Eg, Bans, credits, notes.

MacHac:
I basically copied the commands verbatim from the code that actually handles the weighting.  I can sanitize, but nothing else in the codebase is sanitized.

Edit: It occurs to me that I can't sanitize anything, since the database entries aren't sanitized.  We would need to rebuild our entire database + recode everything in the code referencing SQL or risk players with spaces in their names losing their entries.

I think that the danger is reasonably low that anyone will be able to get a dangerous name, which is probably what the people who wrote every other bit of sql code in the game (Including Kn0ss0s) thought.

DavidEdwards:
Ckeys are safe. This however isn't safe to merge because a !FUN! player can spam the verb and probably crash the server with SQL queries. That needs to be regulated. Once per round is plenty enough. So this could be cached in /client for example.

Iamgoofball:
You should be told this after antags are assigned if you dont get antag instead, or when you connect

DavidEdwards:
@Iamgoofball  Yes, even better. 

The value doesn't change except through the antag selection process.

@MacHac You could edit my antag change to set an antag weight value in the /client. Do **not** print the value out everytime it changes in the selection process. The user could end up with over 10 lines of antag processing printed out if it is a big round. They will see what kind of round it is before it has even started.

PR #933, opened 2016-04-17
Semi-port drying agent from tg and wet turf reworks.

Nirnael:
-Adds semi-port of drying agent from tg + some constants and turf drying
proc.
-Makes turf slippery when emptying extiguisher

-Skips changelog because github is retarded
~Nirnael


oisin100:
Good job!

Code looks good. Not sure what a semi-port is supposed to be :P

:+1:  x :100:

Nirnael:
Well i changed around some stuff and made it so you're able to spray your galoshes from anywhere, on tg they have to be on the ground. So uhh...I guess its a semi-port. But just port is fine.

ToGWtF:
great work

PR #946, opened 2016-04-18
Shunting no longer adds 30 seconds to nuke timer.

AsV9:
Because it made getting the nuke so fucking difficult and/or made shunting basically useless.
The nuke is tied to the AI core itself so if that dies, the nuke is disabled.


AsV9:
BYOND really hates actually working doesn't it..

ShadowDeath6:
How big of a problem is this? 30 seconds isn't a long time to kill an APC.

X-TheDark:
That was actually 60 real-time seconds, as it added 30 BYOND seconds to the explosion timer.

PR #949, opened 2016-04-18
Admin R_PERMISSIONS job whitelist toggle

DavidEdwards:
Allow upper admins to determine eligibility for players to be whitelisted


Groudonmaster:
Took you long enough :^)

Issue #895, opened 2016-04-14
VIRUSES AREN'T CURABLE AT ALL


ShadowDeath6:
### Expected Behaviour

Cure a virus and develop antibodies so you can no longer get a disease
### Actual Behaviour

Everyone on the station has brain AIDS or fucking skin burn that doesn't go away ever
### Steps to reproduce the problem
1. Get level 7 biohazard virus
2. Cure it
3. It comes back
4. ask admin to cure it
5. IT DOESN'T GO AWAY

If this is a dupe of another issue then fuck me
Critical priority also pls

FluffySurvivor:
Was this fixed?

Nirnael:
I just tested it and it seems to be working fine.

Cruix:
There is probably a runtime in a specific symptom.

Cruix:
And I accidentally hit close and comment

Nirnael:
It seems like whoever createad certain diseases gave them a cure but made them incurable, literally 1 constant is all it took to fuck shit up. But i dunno if this was intended so until it's clear i'm not fixing it.

oisin100:
Fixed in #951

PR #951, opened 2016-04-19
Virus/disease immunity fix and some tweaks.

### LINKED ISSUE #716: Magnitis is completely borked
### Expected Behaviour

Catch virus, cure it with iron, be happy
### Actual Behaviour

Cure virus with iron, catch it again because you don't develop antibodies, then continue to have symptoms after curing it a second time and wearing a biosuit, without the disease showing up on the medscanner

Catch magnitis and cry

Nirnael:
### Fix #716 #895

The way diseases code works is you have curable, resistible and
carriable diseases (ONLY PREMADE DISEASES, HARDCODED, NOT MADE INGAME).
So right now, to have a disease removed from your body, i.e. cured, a
disease must have curable and resistible flags, that is the disease can
be cured and can be resisted after being cured (antibodies or whatever).
If a disease is only curable but not resistible, you can lower the
stage. but not remove it from your system, meaning cure it. Carriable is
useless and I don't know why it's in the code when it's not being used
anywhere.

So, all curable diseases are also necessarily resisted against because
of code (it shouldn't work that way IMO, diseases should probably
evolve, maybe mutadone, etc.). The list **>>OF CHANGED/FIXED DISEASES<<**
is:
**CURABLE & RESISTIBLE:**
-MAGNITIS
-GBS
-SPACE RETROVIRUS
-WIZARDITIS
-ALL TRANSFORMATION VIRUSES EXCEPT ZOMBIE

**INCURABLE AND IRRESISTIBLE:**
-TRANSFORMATION ZOMBIE (once infected, cannot lower stage and cannot
cure, you become a zombie)

Everything else is the same, curable and resistible, or whatever, they
haven't been touched. Imho code should be changed where curable means
actually curable and not "possibly lowering the stage but not actually
curing it" and some other bullshit. I'll do it, but I wanna know if
you're ok with how diseases are right now and if I should change any to
incurable, etc.


FluffySurvivor:
If you're making the zombie virus incurable make antags be immune to be RNGed by the random event.

code/datums/diseases/transformation.dm
```diff
@@ -124,12 +124,12 @@
 	viable_mobtypes = list(/mob/living/carbon/human)
 	permeability_mod = 1
 	cure_chance = 0
```
oisin100, line 126: Give zombies a cure chance. Not being able to cure the rage virus is idiotic. Zombies itself should be in its own gamemode.

oisin100:
Code looks good.

Nirnael:
Well, I don´t know tbh, it was incurable before. But if you want it to be curable, then the question is what chance for cure, and possibly more reagents for cure?? It seems kinda ezpz to get rid of diseases the way they are now. I guess different gamemode works as well, but same question, infection chance, cure, cure chance, etc.

oisin100:
Whiskey, It should be something like 3% chance of cure per tick.
If zombies was a gamemode then i wouldn't mind it incurable as in most movies its not curable anyway.

DavidEdwards:
Zombies were always meant to be curable. I set Whiskey to be its cure at one point, until I added a way for the disease to be researchable. At which point I wanted to make it so that even turned zombies could be (perilously) returned to human-ness. Since these zombies aren't "the undead", they are more rage-afflicted.

Even in gamemode they should be curable because that could be a survivor victory condition.

Nirnael:
Well if we are doing a gamemode then sure. However rage disease was incurable in code before I changed  it. Since it was created, according to history, the cure chance was never changed from 0.

 I'd suggest we make a gamemode where one or various people are immune to it and you would have to test them to synthesize a vaccine from the immune ones/s and inject the infected to revert them back to human. Also, introduce a "hulk" type tanky zombie so that when the survivors turtle with walls around viro, where the vaccine would be made (worst, metagamey possible case), hulks can breach walls and let the other zombies in, together with increased resistance to damage so they don't get shrekt by lasers.

Maybe a low airborne contraction chance so the initial zombies don't get cuffed and spaced in 10 seconds and end the round. Or everybody contracts it, low stage advancement chance, not showing on med analyzer until it's found and researched(basically imo it shouldn't show outright on the med analyzer since it's an unknown disease, you would have to test the blood of infected, use viro to figure out it's zombie and find immunes for vaccine), so that immunes go get the vaccine or everybody works towards it if found out early.
Please tag work in progress.

DavidEdwards:
I had intended to let multiple zombies "activate" the same object to use their combined crowd strength to open things. Like force opening doors after X seconds. 

Zombies to be able to get "points" for eating brains, and spending those on upgrades like spitting the virus (or airborne as you say), hulking up, speed bonuses, etc etc. So more brain food more power.

That is kinda going against the whole 28 days later theme, but I needed some metric to determine upgrades.

Nirnael:
Hmm, maybe random zombie mutations from eating brains as opposed to choosing them? But there's definitely some juice in this and should be made its own gamemode in the style of blob crew vs danger.

Nirnael:
Closed for rework.

PR #892, opened 2016-04-14
Makes slime processor scoop up dead slimes nearby automatically

ghost:
It scoops up dead slimes in a 1 tile range right next to it automatically.

Quality of life changes.

Port from /tg/, credit to goofball.

EDIT by @X-TheDark
Ported slime processor from tgstation/-tg-station#16870 by @Iamgoofball 
This has since been ported manually somehow by Adam.

PR #955, opened 2016-04-19
Ports /tg/ slime processor picking up nearby dead slimes

X-TheDark:
Revisit of #892 

Reimplementation of /tg/ Slime Processor scooping up slimes thing.
Not linking because original credit is in the linked issue.

Changes:
- Uses orange instead of range to get the slimes. As range also processes the center (and it's contents, which would be the scooped up slimes), as per http://www.byond.com/docs/ref/info.html#/proc/range

### DIFF
--- a/code/modules/food&drinks/kitchen machinery/processor.dm
+++ b/code/modules/food&drinks/kitchen machinery/processor.dm
@@ -29,6 +29,23 @@
 	for(var/obj/item/weapon/stock_parts/manipulator/M in component_parts)
 		rating_speed = M.rating
 
+/obj/machinery/processor/process()
+	..()
+	var/mob/living/simple_animal/slime/picked_slime
+	for(var/mob/living/simple_animal/slime/slime in orange(1,src))
+		if(istype(slime, /mob/living/simple_animal/slime))
+			if(slime.stat)
+				picked_slime = slime
+				break
+	if(!picked_slime)
+		return
+	var/datum/food_processor_process/P = select_recipe(picked_slime)
+	if (!P)
+		return
+
+	src.visible_message("[picked_slime] is sucked into [src].")
+	picked_slime.loc = src
+		
 /datum/food_processor_process
 	var/input
 	var/output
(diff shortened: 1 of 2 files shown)


code/modules/food&drinks/kitchen machinery/processor.dm
```diff
@@ -29,6 +29,23 @@
 	for(var/obj/item/weapon/stock_parts/manipulator/M in component_parts)
 		rating_speed = M.rating
 
+/obj/machinery/processor/process()
+	..()
+	var/mob/living/simple_animal/slime/picked_slime
+	for(var/mob/living/simple_animal/slime/slime in orange(1,src))
```
DavidEdwards, line 35: This is running on each process() tick. Is this going to make a :crying_cat_face: ?

X-TheDark, line 35: Well, it's either this or no slime pickup. I'm not sure there's any other way to do this. Feel free to not merge it, though, as I understand the processing concerns.

Maybe there's something clever one can do with checking for nearby processors when the slime corpse is dragged or something.

ghost:
if(slime.loc == src)
    return
should be

if(slime.loc == src)
    continue

also, thanks @Iamgoofball for the actual code-credited the wrong guy originally

code/modules/food&drinks/kitchen machinery/processor.dm
X-TheDark, line 35: Actually, I don't think there's a more efficient way to do it. Even if you change this and somehow move it to process when the dead slime is moved, you still end up with a orange() check for nearby processors, which means it's going to be ran X (where X is the number of slimes (since there's no subtype for dead mobs)) times, instead of once per tick.

I am open to suggestions on how this can be implemented in a better way, though.

X-TheDark:
I use orange() instead of range() so that part of code is not needed anymore.

code/modules/food&drinks/kitchen machinery/processor.dm
Iamgoofball, line 35: This is fine.

PR #961, opened 2016-04-19
Buffs Cleanbot AI

### LINKED ISSUE #960: Cleanbots stack on the same messes
### Expected Behaviour

Release a swarm of 6 cleanbots on patrol mode, the station is spotless.
### Actual Behaviour

All 6 bots stack on the same tile and target the same messes, you may as well only have 1 cleanbot.
### Solution

Merge PR #961

Cruix:
Fixes #960 

Makes cleanbots avoid targeting the same mess as other cleanbots.

### DIFF
--- a/code/game/machinery/bots/cleanbot.dm
+++ b/code/game/machinery/bots/cleanbot.dm
@@ -130,6 +130,9 @@ text("<A href='?src=\ref[src];power=1'>[on ? "On" : "Off"]</A>"))
 /obj/machinery/bot/cleanbot/process_scan(obj/effect/decal/cleanable/D)
 	for(var/T in target_types)
 		if(istype(D, T))
+			for(var/obj/machinery/bot/cleanbot/other_bot in SSbot.processing)
+				if(D == other_bot.target)
+					return null
 			return D
 
 /obj/machinery/bot/cleanbot/bot_process()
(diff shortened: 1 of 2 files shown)


ghost:
can you port /tg/'s floorbots

code/game/machinery/bots/cleanbot.dm
```diff
@@ -130,6 +130,9 @@ text("<A href='?src=\ref[src];power=1'>[on ? "On" : "Off"]</A>"))
 /obj/machinery/bot/cleanbot/process_scan(obj/effect/decal/cleanable/D)
 	for(var/T in target_types)
 		if(istype(D, T))
+			for(var/obj/machinery/bot/cleanbot/other_bot in SSbot.processing)
+				if(D == other_bot.target)
+					return null
```
X-TheDark, line 135: Shouldn't this be break, so it breaks out of the cleanbot loop and continues scanning the target_types.

I think that right now, if there's a bot targeting a square, the other bots will become stuck when they process it, because they return, rather than trying to pick another target from target_types.

Cruix, line 135: No, this proc only processes one item at a time, the loop for all objects in view is in the bot/scan() proc which calls this one. This proc returns null when the target is invalid, and the target when it is valid.

Cruix:
Yes, I can port the floorbots.

Issue #886, opened 2016-04-13
A not shit traitor item


ghost:
Pickpocket gloves.

Cost: 2 tc.

Sprite- black gloves.

Description- 'A pair of sticky, shiny black gloves.'

Detective fibre scanner display-'fibres from a shiny glove.'

Use-

Multiple.

They provide no insulation from shock, but work as regular black gloves otherwise-even black gloves reduce shock damage.

In the strip menu, they allow you to see the contents of belts and backpacks before you pull them off. You can also directly remove items from belts and backpacks, even if they're on someone else.

Items pickpocketed go directly into the inhand, not on the floor.

You are able to see what is in someone's pockets in the strip menu.

Pickpocket speed is vastly increased-removing id's or items from pockets takes about the same time as it does that slipping someone on soap stuns them. 

Removing items from backpacks or belts takes 3 seconds.

All of these are as silent as regular pickpocketing-the person being stolen from will not notice unless they move, at which point they will feel their pockets, id, or backpack/belt/anything in suit storage (this includes tazers in an officer's armor, the cmo's hypo on his labcoat) be 'nimbly grasped at'.

Suit storage stripping takes normal strip time but is otherwise silent.

Strip time all around is reduced by 33%. For anything.

The gloves also allow you to directly place items in belts/backpacks/pockets with increased speed.

Now, remember all of this requires you and the person to be standing still. If they so much as fart they'll notice and you're sailing out the airlock.

This offers both great potential in theft, or framing (hide an emag in your target's backpack and scream for sec to arrest him, or assasination (hide a grenade in your target's pocket/steal the oxygen tank right off the ce's hardsuit in the middle of space).

It also allows for more amusing shit, like slipping the captain, tearing his jumpsuit off and filling his backpack with pictures of your ass before jumping into disposals leaving him confused, humiliated and naked.

Remember all of this is only good if someone's standing still and they'll notice very fast you just stole something from them when it's gone after you slip them.

Cruix:
> Be atmos tech
> Get tank full of plasma/N2O mix
> Place in target pocket
> toggle target's internals
> laugh manically

ghost:
exactly

ShadowDeath6:
You actually can't toggle on internals unless the item is in exosuit storage.

Super3222:
Riot shields should be a hard counter for this.

ShadowDeath6:
Close this >:^(

PR #974, opened 2016-04-21
Pickpocket gloves (and a whole baggage of reworks)

X-TheDark:
Refer to issue #886.

**TL:DR Pickpocket Gloves + alot of code to support that : faster stripping, can plant/strip from backpack/belt. Also can see pockets/toggle internals from pocket slots. Masks can now be adjusted on person, regardless of these gloves. Instant, but visible without the special gloves.**

**Probably needs some testing too, but I've tested it with probably most common human interactions during implementation/testing. Other mob stripping will probably not work as well.**
### Gloves themselves

Pickpocket gloves, as Adam envisioned, with an additional touch from me by making the Chameleon.

Features of gloves themselves:
- New subtype of gloves (/gloves/pickpocket) that have 3 visibility flags which control what extra slots you see or don't see.;
- **Strip coefficient** that controls how much faster stripping/planting happens (almost all of it). Ranges up to 1 (0 = 0% faster, 1 = 100% faster, aka instant, supports negative values);
- The only exception to the above is stripping/planting container items directly - stripping from container is full-container time (explained below), planting into container (backpack/belt, btw, no planting insider containers inside containers) respects strip coefficient (so it's faster to plant);
- Containers now have a **strip_time** var defined - tells you how long it takes to STRIP from this container, when it's on the person (only applicable to backpacks/belts at the moment, but defined for all storage objects). PLANTING is multiplied by strip coefficient, making it faster (barring negative coefficient value shenanigans);
- Stipping/Planting is silent to the target;
- All movement rules apply normally - they move, you lose;
- If pockets contain internal tanks, you can switch them on/off now;
- Stripping puts the item into your hand. For stripping items, if they cannot be placed into your ACTIVE HAND, they get dropped UNDER YOUR MOB (instead of under your target)
- You can only strip one item from container (backpack/belt on target) at a time, due to how stripping/planting is implemented (read in the instruction book below);
- Can silently toggle internals (including Pocket slots);
### The enclosed instruction book:
- Stripping normal slots works the same way;
- To PLANT an item into a container, hold item in your hand and CLICK ON ONE OF THE CONTENTS OF THE BACKPACK (or the word Empty, if it's empty);
- To STRIP an item from a container, have YOUR HAND EMPTY and click on the item in the container;
- Available for 4 TC from your nearest Syndicate Uplink;
## Groundwork changes (reworks)
- obj/clothing variable **alt_desc** changed to **fiber_text** (as it was previously used **IN 2 ITEMS OUT OF ALL THE CLOTHES EXISTING**, namely syndie hardsuit/helmet. Alt_desc defined for those two separately);
- As per above, you now leave **fiber_text** fibertext on anything you touch, this allows different item names and fibers. By default, **fiber_text is set to name of item (in New())**;
- Detective scanner, obviously, now uses fiber_text;
- **New href var** used in show_inv() for humans - **item_ref**, which saves the reference of an item in the container OR 0, if it's empty;
- New slot - **slot_in_belt** : refers to item in, well, belt;
- Various mob/species inventory procs (can_equip/get_item_by_slot) support slot_in_belt, as well as call parent functions and process only their own slots. Additionally, get_item_by_slot now has a **new optional parameter of type /obj/item**, which, if defined, checks that the item is inside backpack/belt (depending on the slot specified). If the optional parameter is not set, it returns the back/belt slots.
- stripPanelEquip/stripPanelUnequip reworked to be capable of working with items inside backpacks/belts.
- Probably some more things I cannot remember...

Oh, right, also, new feature - you can now slide masks off/on someone. It takes 2 seconds to do and is instant with THE GLOVES. People around are alerted when someone does this to you, if you do it without them (similar to taking something off someone - big shiny red message).
# New Inventory Screen

new_inventory

:cl: X-TheDark
rscadd: Syndicate agents can now purchase Pickpocket Gloves from Stealthy Tools section for 4 TC. These gloves can disguise themselves as other gloves by holding them in-hand and clicking on them.
rscadd: Pickpocket gloves make stripping/planting_on_person 33% faster than normal. You can also remove/place items directly into belts/backpacks worn by the target. These gloves also allow you to see what's inside the targets pockets. In addition, stripping/planting items is SILENT with these (target only gets the message of the attempt if you fail). As a bonus, the item you strip from the target is placed in your hand, instead of the targets feet.
rscadd: To place an item into belt/backpack, hold whatever you want to place in your hand and click on any item in the backpack/belt contents thing. To remove an item, have your hand empty and click on any contents of the backpack/belt.
rscadd: Oh, also, toggling internals on the target is SILENT with these gloves and you can also toggle internals if the target has a emergency tank in one of the pockets.
rscadd: You can now adjust the mask on the target (Slide it in place/off of face), it takes 2 seconds. With pickpocket gloves, it's instant.
/:cl:

### DIFF
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -445,6 +445,12 @@ var/list/uplink_items = list()
 	item = /obj/item/clothing/under/chameleon
 	cost = 2
 
+/datum/uplink_item/stealthy_tools/chameleon_pickgloves
+	name = "Chameleon Pickpocketing Gloves"
+	desc = "A pair of gloves that make pickpocketing silent, speeds up stripping/planting actions (33%) and allows extra pickpocketing options. Cam camouflage as other types of gloves. Leaves very specific fibers that make it easy to tell these were used if found by detective."
+	item = /obj/item/clothing/gloves/pickpocket/chameleon
+	cost = 4
+
 /datum/uplink_item/stealthy_tools/chameleon_stamp
 	name = "Chameleon Stamp"
 	desc = "A stamp that can be activated to imitate an official Nanotrasen Stamp. The disguised stamp will work exactly like the real stamp and will allow you to forge false documents to gain access or equipment; \
--- a/code/datums/uplink_item.dm
+++ b/code/datums/uplink_item.dm
@@ -447,7 +447,7 @@ var/list/uplink_items = list()
 
 /datum/uplink_item/stealthy_tools/chameleon_pickgloves
 	name = "Chameleon Pickpocketing Gloves"
-	desc = "A pair of gloves that make pickpocketing silent, speeds up stripping/planting actions (33%) and allows extra pickpocketing options. Cam camouflage as other types of gloves. Leaves very specific fibers that make it easy to tell these were used if found by detective."
+	desc = "A pair of gloves that make pickpocketing silent, speeds up stripping/planting actions (33%) and allows extra pickpocketing options. Can camouflage as other types of gloves. Leaves very specific fibers that make it easy to tell these were used if found by detective."
 	item = /obj/item/clothing/gloves/pickpocket/chameleon
 	cost = 4
 
--- a/code/__DEFINES/clothing.dm
+++ b/code/__DEFINES/clothing.dm
@@ -44,6 +44,7 @@
 #define slot_s_store		17
 #define slot_in_backpack	18
 #define slot_legcuffed		19
+#define slot_in_belt		20
 
 //Cant seem to find a mob bitflags area other than the powers one
 
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -24,6 +24,7 @@
 	var/allow_quick_gather	//Set this variable to allow the object to have the 'toggle mode' verb, which quickly collects all items from a tile.
 	var/collection_mode = 1;  //0 = pick one at a time, 1 = pick all on tile, 2 = pick all of a type
 	var/preposition = "in" // You put things 'in' a bag, but trays need 'on'.
+	var/strip_time = 30 //Time it takes to directly strip/put-things-into the container from from the strip menu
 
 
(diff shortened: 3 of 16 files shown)


ToGWtF:
Jesus fuck this is the greatest thing i've seen.

FluffySurvivor:
I respect this PR.

code/datums/uplink_item.dm
```diff
@@ -445,6 +445,12 @@ var/list/uplink_items = list()
 	item = /obj/item/clothing/under/chameleon
 	cost = 2
 
+/datum/uplink_item/stealthy_tools/chameleon_pickgloves
+	name = "Chameleon Pickpocketing Gloves"
+	desc = "A pair of gloves that make pickpocketing silent, speeds up stripping/planting actions (33%) and allows extra pickpocketing options. Cam camouflage as other types of gloves. Leaves very specific fibers that make it easy to tell these were used if found by detective."
```
oisin100: Can*

AsV9:
> be in n2o gas
> grayshirt mcfuckshit comes along
> insta slides off mask
> cry

Super3222:
This is awesome.

X-TheDark:
> be in n2o gas
> grayshirt mcfuckshit comes along
> insta slides off mask

Fair enough, fair enough. I was hoping to make this a overall mechanic, but I agree that having your mask slid off instantly by a random greyshirt would probably suck. I'll change it so that if you have the gloves (more specifically, silent internals-enabled gloves), you'll slide the mask instantly.

X-TheDark:
Changes made:
- Sliding Mask On/Off from strip panel takes 2 seconds, unless you have THE GLOVES, then it's instant.
- Slight changes to how mask_adjust variable works. Makes it trigger my OCD less.
- Makes the Slide On/Off button not appear on masks that cannot be slid on/off (gasmasks only, it seems).

Groudonmaster:
I respect this PR and I respect you Dark. This is impressive.

ghost:
so when is dark the fucking headcoder

PR #982, opened 2016-04-21
ports changelingrain's removal of EMP cap size

ghost:
well uh

the title explains it pretty well

ports /tg/'s removal of EMP cap size

300u iron 300u uranium bluespace grenades will probably send the station back to the stone age with this

:cl: Joan
rscadd: Removes EMP effect cap size. Ported from /tg/ by Adam El-Tablawy.
/:cl:

### DIFF
--- a/code/game/objects/empulse.dm
+++ b/code/game/objects/empulse.dm
@@ -20,7 +20,8 @@
 	if(heavy_range > light_range)
 		light_range = heavy_range
 
-	for(var/atom/T in range(light_range, epicenter))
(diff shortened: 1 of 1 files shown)


oisin100:
Me likey.

Changelog pliz.

And have you tested this? The code doesent look like it changed anything regarding EMP capsize(And are you sure we had a EMP capsize in the first place?)

ghost:
was tested

it handles the same was as an explosion now rather than dying out past an arbitrary range

Issue #977, opened 2016-04-21
powersinks suck ass


ghost:
### Expected Behaviour

it actually drains enough power to shut off lights, equipment, ect, and is a serious inconvenience
### Actual Behaviour

sometimes it depowers the central hallway vending machines and nothing else
### Steps to reproduce the problem

1.buy powersink
2. attatch powersink
3. it powerstinks.

seriously you can't even use it if someone set up the tesla or sing and even if you decon the fucking smes it still won't drain power

increase powerdrain by a fuckton and give it a bigger explosion capacity pls

oisin100:
It blows up before it does anything usefull :(

ShadowDeath6:
It's stupidly powerful if you use it before engine is up, especially if you decon the smes
The problem is that people use it when smes are fully charged

PR #986, opened 2016-04-21
Massively buffs powersink

X-TheDark:
Refer to issue #977.

**If you think this is too heavyhanded, feel free to close this, I'll then see about making a powersink that takes into account the power output of whatever singularity the station is using.**

Now that we have the tesla, powersink just plain sucks ass, if the tesla is set up. If it's not, why are you buying this in the first place, I wonder?

Changes:
- Powersink realistically drains ALL produced power @ 8 MW of power per tick (up from 0.6 MW);
- Powersink has a 4 times greater capacity @ 400 000 000 (up from 100 000 000);
### NOW, BEFORE YOU CALL ME OUT OF MY GODDAMN MIND, please read this:
- Once the tesla is setup, if it is wired directly into the grid, it produces 2 MW of energy (which gets to about 3+ MW in the later stages, which might still sometimes invalidate this device);
- Old singularity gives out 2.5 MW when the collectors are full and it is at Stage 2-3 thing (The stable place it stays with PA 0, basically);

Let that sink in for a moment...a 5 minute effort of turning on emitters/shield gens (no modification to layout) and a minor rewiring is getting you GODDAMN 2 MW OF POWER PER TICK, which increases as time goes on to 3+MW.

At this point, a 600kW drain Powersink is not going to cut it at all, hell, it doesn't even drain max output SMES (800k total).

P.S. As a minor thing. This is actually impossible to balance, as, if the singularity (any kind) IS NOT wired to the grid, you get 0.8 MW power, which is going to make this puppy run for a LOOONG time.

Now, if the singularity IS WIRED to the grid, **EVEN WITH THESE RIDICULOUS NUMBERS, you are looking at about a 5 minute powersink life time, out of which, about 1:50 minutes are used to drain the power from station's APCs first (I tested this myself)**. After that you get about 3:34 minutes of blackout.

This can be mitigated by using the old singularity in a risky way - if you keep it around Stage 4 (7x7 size - fills entire containment), it will produce 5 MW (when it's just become S4) up to about 6-7 MW, before you need to turn power off (or not, I won't judge). This reduces the blackout time to about 40 seconds.

:cl: X-TheDark
rscadd: Powersink severely buffed. You can no longer hope to outproduce it's consumption ability. Powersink drains up to 8 MW per second (up from 0.6 MW) and it's capacity is increased from 100 to 400 MW.
rscadd: AI cries, Powernet dies. If you don't have the most basic power set up, you can kiss your power goodbye for the entire round, unless you find this thing.
rscadd: Careful maintenance of high-stage singularity (Stage 4 - the max containment size one) powered production will DRASTICALLY reduce the powersink lifetime, due to ridiculous amounts of energy it produces. Makes sure to plug it into the grid! And makes sure it doesn't grow any more.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -12,9 +12,9 @@
 	throw_range = 2
 	materials = list(MAT_METAL=750)
 	origin_tech = "powerstorage=3;syndicate=5"
-	var/drain_rate = 3000000		// amount of power to drain per tick
+	var/drain_rate = 8000000		// amount of power to drain per tick
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 3e8		// maximum power that can be drained before exploding
+	var/max_power = 4e8		// maximum power that can be drained before exploding
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = 0 // stop spam, only warn the admins once that we are about to boom
 
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -14,7 +14,7 @@
 	origin_tech = "powerstorage=3;syndicate=5"
 	var/drain_rate = 3000000		// amount of power to drain per tick
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 6e8		// maximum power that can be drained before exploding
+	var/max_power = 3e8		// maximum power that can be drained before exploding
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = 0 // stop spam, only warn the admins once that we are about to boom
 
--- a/code/game/objects/items/devices/powersink.dm
+++ b/code/game/objects/items/devices/powersink.dm
@@ -12,9 +12,9 @@
 	throw_range = 2
 	materials = list(MAT_METAL=750)
 	origin_tech = "powerstorage=3;syndicate=5"
-	var/drain_rate = 600000		// amount of power to drain per tick
+	var/drain_rate = 3000000		// amount of power to drain per tick
 	var/power_drained = 0 		// has drained this much power
-	var/max_power = 1e8		// maximum power that can be drained before exploding
+	var/max_power = 6e8		// maximum power that can be drained before exploding
 	var/mode = 0		// 0 = off, 1=clamped (off), 2=operating
 	var/admins_warned = 0 // stop spam, only warn the admins once that we are about to boom
 


ghost:
not to mention engineering insists on making like 3 teslas and a sing 

also, you may want to slightly increase powersink cost if it can run for 30 minutes with nobody hunting because sometimes traitors might make cheese sinks that lead into some hidden ass corner in maint in a wall

X-TheDark:
This would only be a problem if singularity (any) is not directly connected to the grid. As I mentioned, I have no idea how to balance this.

That being said, maybe cost increase is the way to go, but how much. It already costs 10 TC, making it pretty much infeasible to buy anything else. Making it costs 12 TC makes emag out of your reach and that is way too good to pass up.

Either this is so strong that without connecting sing to grid you are literally powerless for the entire round or it's weak enough to drain SMES-connected powergrid, but not even close to enough to drain grid-connected powergrid.

X-TheDark:
Changes in last commit:
- Halves the capacity (300 MW instead of 600 MW)

After deliberating for a bit, I think that a 4.5 minutes (approx) total blackout during a grid-connected tesla is long enough for the traitor to do whatever he wanted to do. And if it's not a grid-connected singularity, that's about 15-20 minutes of no power.

ShadowDeath6:
How is this versus OG (good) singularity? If this encourages engineers to use tesla every single round in case of a powersink it would make me salty.

X-TheDark:
Lemme check real quick.
S3 Singulo (5x5) = ~2.5 MW at max plasma canisters in collectors at the stable PA set to 0 point. (Goddamn, this thing fluctuates as much as Tesla does - multitool, SMES, all show different things. Going with multitool readings)

And the powersink lasts about the same, actually, about 4.5 minutes.

X-TheDark:
**Further testing**
A newly created S4 Singularity (7x7 - max containment size) gives about 4.5 MW of power, this is much, MUCH more than the Tesla did.

The current (buffed) Powersinks last about 2.5-3 minutes under those conditions, without even draining any power and causing a blackout.

Even with a buffed drain rate of 8 MW, it didn't manage to drain the power from a full powered station with S4 singularity before blowing up (to be more exact, it blew up in 2 minutes, 1:50 of which was spent emptying the power)...so yeah, even under these conditions, the powersink sucks ass.
- Going to buff the drain rate and the max_charge rate.

X-TheDark:
Ok, after myriad of tests, I decided on these values:
- Drain speed = 8 MW/sec. You cannot outproduce it, you can only hope to feed it as fast as possible to blow it up;
- Max capacity = 400 MW. In my tests, the "regular" setup (tesla/PA 0 singularity + full plasma cans) produce enough power for the sink to live about 5 minutes. 1:50 are spent draining the station's full APC from 100-0 and about 3:34 is the blackout time (3.5 minutes). If you are running a high-stage singularity (Stage 4, the one that fills the containment), it produces so much power that the blackout time becomes 40-60 seconds.

DavidEdwards:
Lets give it a go. Always easily revertable.

X-TheDark:
Death to incompetent engineers!

PR #988, opened 2016-04-22
Nerfs Lich

### LINKED ISSUE #976: Lich Wizard Phylacteries are Immpossible to Find
### Expected Behaviour

There are counters to Lich wizards.
### Actual Behaviour

Cruix goes lich every time he rolls wizard and gets outrobusted by the crew at least 5 times, but still lives because his phylactery is in a 1x1 room you can only get into by ethereal jaunting.
### Solution

Make the pinpointer locate phylacteries.

or

Make using ARISE!!! drop an item on the lich's old body that points toward the phylactery, possibly deactivating when the ability is off cooldown, to give the wizard a chance to hide it again after a few minutes.
EDIT:
or

There is a revenant-style beam from the old body to the new body for around 10 seconds after the lich "dies".

Cruix:
Fixes #976 

Ports /tg/'s beam from a lich's old dead body to their new one when the lich revives. The beam lasts 15 seconds, about as long as the lich is stunned for.


Cruix:
Why did it delete the entire revenant file?

Cruix:
Leave a comment

Issue #975, opened 2016-04-21
Proposed race changes to make races more complex and appealing


ghost:
### Phytosian changes

The issue with phytosians is that their slow heal-While extremely slow-is still infinitely favorable to playing as a regular, boring human, especially considering not many people exploit the high-damage-burn-shit on phytosians.

Phytosians need more severe downsides and upsides, in all honesty, to encourage people playing different races aside from cabbages.

Here are a list of proposed changes:

The burn damage is not currently a flat 2x. Make it as such, so that a welder will deal 30 rather than 15, a laser will deal 40 rather than 32, et cetera.

Make phytosians heal from certain gardening chems. The specifics-

Saltpetre heals 2.5 burn and 1 tox per tick.

Ammonia heals 1 burn and 1 brute per tick.

Robust Harvest heals 2 tox per tick, but slows down the metabolization of harmful chemicals in the body.

Left 4 Zed functions like a weak unstable mutagen-Causing a low chance of random (usually harmful) mutations, healing 1 burn, and causing one tox per tick.

E-Z Nutrient heals 1 tox and 4 oxy per tick, and stabilizes at -50 health, preventing any further oxygen loss. It is also addictive in quantities higher than 15u and withdrawal adds 2 oxy and +.5 losebreath a tick.

Plant-B-Gone (Weedkiller) causes 3 toxin and +.5 losebreath per tick, in addition to causing confusion (a la cryptobolin.)

Diethylamine- Heals .5 burn, .5 brute, and 2 toxin and 2 oxy per tick, in addition to curing hallucinations like synaptizine. OD's at 20u and causes rapid toxin and brain damage.

Ethanol (or any alchohol drink) -Causes 2 brain damage and .4 toxin damage a tick, in addition to confusion. Alchohol is not good for plants! Ingesting more than 30u of any drink with a strength stronger than 2 (second highest) or pure ethanol causes heavy toxin damage and knocks you out in 50 cycles.

Sugar-Increases the healing rate of phytosians as long as it is in the body to twice it's normal strength, but phytosians have lower tolerance for sugar and enter hyperglaucomic shock (which already exists in game, sugar comas are fucking nasty) in quantities above 40u. In addition, you also take much more toxin damage in darkness.

The floral somatoray:

The floral somatoray is an R&D counter to rampaging phytosians in much the same way the tempgun is to slimes. The yield setting does nothing to phytosians-The 'Mutate' setting functions exactly like a decloner, but weaker.

[...]

PR #990, opened 2016-04-22
Splits Languages var into Languages Understood and Languages Spoken

Cruix:
Possibly important for #975 

Splits the languages var into languages_understood and languages_spoken, allowing atoms to understand specific languages without speaking them.


code/modules/mob/living/silicon/silicon.dm
```diff
@@ -1,7 +1,8 @@
 /mob/living/silicon
 	gender = NEUTER
 	voice_name = "synthesized voice"
-	languages = ROBOT | HUMAN
+	languages_understood = ROBOT | HUMAN
+	languages_spoken = HUMAN
```
X-TheDark: Will this interfere with silicon ability to use the binary channel?

Cruix: No, I have tested and the binary channel works fine. As far as I can find, nothing in the game currently speaks robot without also speaking human. Actually I'm not sure why I left it like this, I suppose I just missed it while I was going through all the languages vars.

PR #994, opened 2016-04-22
admin spawned temperature gun made not shit

ToGWtF:
because abuse


oisin100:
10/10 Most descriptive commit name ever.

Super3222:
give the players the shitty version
but the admins get the actually good version

also that commit will get your ass ridden on by some old coders

DavidEdwards:
> also that commit will get your ass ridden on by some old coders

There are shitty commit names and there are really shitty commit names. :D 

In the future, nobody will know wtf that commit is referencing.

@ToGWtF lets re-do it?

Issue #847, opened 2016-04-09
Operating Computer Ideas


Super3222:
So let's refer to #10, which suggests that people who are laid out on the surgery table's implants should show up on a surgery computer around it. 

Implants like the syndicates own will show up as "#ERROR IMPLANT NOT RECOGNISED#" and whatever. 

That's a pretty good idea, and I was planning on working it at one point in time, but still far too easy.

So currently the operating computer's functions is to tell you the vitals of the patient on the surgery table as well as the next step in the surgery you're conducting (which literally no one checks for).

So that's why I wanted to suggest we add a big book in surgery.
One called "Surgery for dummies: a guide that teaches you more to surgery other than conducting sex changes".
It'd realistically, link to the wikia's page on surgery. And some information on the operation table.

So  operating tables and computers should be able to be researched in R&D. That's a given, and would most likely require a rollerbed upon construction.

But what I wanted to get at is more functions being given to the operating computer.
I want it to be able to detect different implants in people as well as different organs, creating a list of them.

However to foreign implants/organs, it won't be able to notice them.

I was thinking the operating computer could have a mode, where it switches to a page with two slots, that will either accept implants/organs. The computer will be be able to scan them, save it to the widespread database of organs that will be previously saved.

And when switching back to the patient mode, the computer would be able to notice the organ that is inside of the patient, and list it.

So, for these implants/organs to be scanned they will need to be placed in a container. Because you can't just slap a xenomorphs brain up against a computer and expect it to be downloaded instantly.

Implants already have their implant cases, so that's pretty handled already.
Organs on the other hand, will be tricky. The computer will not accept spoiled organs, so that means you will have to keep it within a freezer. But how would you be able to conceal it?

That's why I was thinking of an implementation of mini-coolers for medical doctors / surgeons, to conceal ONE organ at a time. (Though, it may have to come with a bag of ice inside of it, running processes of cooling down the objects inside of there.)

 And any other small item they may want to. It'd be a bulky item, like a toolbox, but less space.

The mini-cooler would run a similar process to the freezers, where they'd consistently keep organs at a low temperature preventing them from being spoiled.

The mini-cooler may actually not look like a cooler, but a cubicle like object.

**tldr**:

*operating computers list the organs/implants (that are in their database) inside of patients on the surgery table
*operating computers can scan organs/implants that aren't in their database
*mini-cooler implementation with a bag of ice to cool down organs inside of it.

ShadowDeath6:
[...]

PR #1002, opened 2016-04-23
Operating Computer Changes

Super3222:
Refer to issue #847

Implements a new system for the operating table. Here's a brief rundown, don't ask for a tldr.

Operating Tables will now be able to detect the different organs and implants within a person. However, it will only be able to identify organs/implants that are within their database.

organs
implants

You can add implants and organs by putting them in their case/cubicles.

Implants come from implant cases, so you can put one inside of the computer (while the implant is in the case), scan it, and if it's not already in the database it will upload it to the database.

Organs will need to be put inside of mini coolers/organ freezers/cubicles, which have a device inside of them that prevents them from decaying. However storing them in a freezer altogether would be a preferred choice, since the cubicles are not as strong as them. Three of them start in the surgery room.

The device that prevents the organs from decaying runs on high-capacity cells, and more upgraded versions. You will need medical access to start it up. My suggestion is to capture organs from people's bodies, immediately catch it in your pokeb- I mean your cubicle, and than scan it, etc etc. Don't waste the battery.

You are also able to visit the database, to check out what was recently added / all of the implants you've caught. It'll upload them by name, rather than icon, because I don't want you downloading so much of them unless you HAVE TO (though I could be underestimating byond's cache). It requires medical access though.

Also a new manual, or you may know them as books that link you to wiki pages, has been added to surgery. It will link you to tg's surgery page. I would love to link it to our own wiki page, but we don't have one, unless the wikimods try to "update" the wiki.

Some examples of what the organ-implant database starts with:
-Loyalty Implant
-Chem Implant
-Tracking Implant

Oh, and this last one.
alien

:cl: Super3222
rscadd: Nanotrasen has upgraded their operating computers, allowing them to detect organs and implants inside of their patients. As long as those organs/implants are inside of the operating tables database.
rscadd: Nanotrasen has been able to fund a useful invention known as medical cubicles, which are able to store organs.
rscadd: Nanotrasen's new operating computers are able to scan the implants in implant cases and organs inside of cubicles to update the database.
rscadd: Nanotrasen has been provided a new manual from the United Earth Surgical Association, which explains surgical procedures for dummies.
/:cl:


X-TheDark:
Damn, really nice work!

DavidEdwards:
Brilliant work. Lots of new code here. Someone please check this over. I will merge it onto dev so it can be tested locally.

PR #1005, opened 2016-04-23
Fixes my fuck up with the tickets.

oisin100:
Not sure how i managed to fuck up so bad.

Doesn't really need a changelog, bug has only been there since yesterday.


Groudonmaster:
Nice :+1:

PR #822, opened 2016-04-07
RE: :^) face on mediborg

Super3222:
### Intent of your Pull Request
#746 - Request.
#774 - Former PR

Adds an icon-statey thing to the mediborg, so when you transform into a mediborg, you get to select icon states, and "Smile Screen" will pop up, you get a mediborg icon with a :^) face on it.

PR #1006, opened 2016-04-23
Gives the "smilescreen" option an animation.

Super3222:
#822

It was missing an animation, so now I'm giving it one. Additionally with lights, and a death icon.

:cl:Super3222
bugfix: Choosing the smilescreen option for medical borgs will give you an animation.
rscadd: Smilescreen option will get a special deathoverlay.
/:cl:


Super3222:
Need to make another commit before anyone merges this.

PR #1007, opened 2016-04-23
Warden gets a sechailer in his box.

Super3222:

Warden will start with a sechailer in his box now, instead of a normal breath mask. I'm not sure why this wasn't done before.

So Detective & Lawyers should be the only two jobs of security that don't get sec hailers.

Another thing I wanted to do was give sechailers an action button to adjust them, but there's a hotkey for that.

:cl:Super3222
tweak: Warden now starts with a proper sec hailer.
/:cl:


ToGWtF:
sechailers need screwdrivers to modify them i thought

PR #1010, opened 2016-04-23
Syndie MMIs are now compatible with AI cores

FluffySurvivor:
https://forums.yogstation.net/index.php?threads/make-syndicate-mmi-capable-of-fitting-into-ai-cores.8475/

Will hopefully make the surgery bag purchase not useless.

:cl: FluffySurvivor
tweak: Syndicate MMIs now fit into AI core frames.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/brain/MMI.dm
+++ b/code/modules/mob/living/carbon/brain/MMI.dm
@@ -165,5 +165,5 @@
 
 /obj/item/device/mmi/syndie
 	name = "Syndicate Man-Machine Interface"
-	desc = "Syndicate's own brand of MMI. It enforces laws designed to help Syndicate agents achieve their goals upon cyborgs created with it, can fit into Nanotrasen AI cores as well."
+	desc = "Syndicate's own brand of MMI. It enforces laws designed to help Syndicate agents achieve their goals upon cyborgs created with it and can fit into Nanotrasen AI cores as well."
 	syndiemmi = 1
--- a/code/modules/mob/living/carbon/brain/MMI.dm
+++ b/code/modules/mob/living/carbon/brain/MMI.dm
@@ -165,5 +165,5 @@
 
 /obj/item/device/mmi/syndie
 	name = "Syndicate Man-Machine Interface"
-	desc = "Syndicate's own brand of MMI. It enforces laws designed to help Syndicate agents achieve their goals upon cyborgs created with it, but doesn't fit in Nanotrasen AI cores."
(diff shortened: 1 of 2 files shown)


ghost:
If you wanna make the bag not useless, make the bottle oldchloral or neurotoxin or pancuronium instead of morphine.

code/modules/mob/living/carbon/brain/MMI.dm
```diff
@@ -165,5 +165,5 @@
 
 /obj/item/device/mmi/syndie
 	name = "Syndicate Man-Machine Interface"
-	desc = "Syndicate's own brand of MMI. It enforces laws designed to help Syndicate agents achieve their goals upon cyborgs created with it, but doesn't fit in Nanotrasen AI cores."
+	desc = "Syndicate's own brand of MMI. It enforces laws designed to help Syndicate agents achieve their goals upon cyborgs created with it, can fit into Nanotrasen AI cores as well."
```
Cruix: And can fit into

X-TheDark:
What about Syndicate MMI's allowing the borg to select Syndicate Cyborg module. Would that make the bag not useless?

FluffySurvivor:
The Syndieborg is worth 80 TC. Spending 4 tc to get a 80 TC death machine. 

In practice, it's an extremely powerful murderbone asset. Indeed, the whole process of debraining someone is risky, but it's easily offset when you get the syndieborg and stomp everyone.

ghost:
Here's an idea.

Replace the syndicate surgery bag with a Syndicate Cyborg Frame and a standard MMI, in a box.

The Frame looks exactly like a normal cyborg frame, but on examination has the word 'SYNDICATE' stenciled into the top of it's head.

Upon inserting the brain, the cyborg gains standard emagged borg laws and emagged borg modules.

Basically, it means you don't have to break into robotics to make use of the item-it comes with the MMI and the cyborg frame, now your only difficulty is killing and debraining the guy.

X-TheDark:
Should it be exoskeleton+parts (so you can assemble it when you want) or should it be the full frame (that cannot be put into backpack). Plus, the words SYNDICATE also appear on examination even after the borg is assembled and MMI put into it, correct?

EDIT: Either way, the rework is something that should be discussed separately and not in this PR, I think. Nothing against this fitting into AI.

ghost:
no, not on examine after assemble

ShadowDeath6:
Does the syndie mmi actually do things or is it just a renamed MMI?

FluffySurvivor:
It inserts syndie laws.

DavidEdwards:
Any additional comments?

FluffySurvivor:
I'd say it's good to go

PR #1015, opened 2016-04-24
Various balance changes for cybermen

MacHac:
### Intent of Pull Request

This PR includes some minor changes to the cyberman game mode.
- Cybermen are significantly more vulnerable to EMP grenades.
- De-cybermanization surgery makes more sense in the context of other, similar operations
- De-cybermanization surgery is no longer invariably fatal for the victim

Also,

### DIFF
--- a/code/game/gamemodes/cybermen/cybermen.dm
+++ b/code/game/gamemodes/cybermen/cybermen.dm
@@ -475,8 +475,8 @@ datum/game_mode/proc/update_cybermen_icons_remove(datum/mind/cyberman)
 	return
 
 /datum/cyberman_datum/proc/emp_act(var/mob/living/carbon/human/mob = src, var/severity)
-	mob.Stun(5)
-	mob.Weaken(5)
+	mob.Stun(5*rand(4,10)*severity)
+	mob.Weaken(5*rand(6,10)*severity)
 	mob.silent = max(60, mob.silent)
 	mob.adjustBrainLoss(40)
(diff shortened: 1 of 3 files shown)


Cruix:
Why is clamping the bleeders before you saw open the bone? That's how it goes in organ manipulation.

Also what was wrong with that shade of red?

ghost:
it looked like bloody diharreah tbh

DavidEdwards:
Why remove a step in the removal process? Do the nerf  changes make sense to you Cruix?

Cruix:
Oh, I thought he was adding a step, now I see he is removing one. I'm not sure why. I'm fine with everything other than the surgery change, but it's not a big deal, since the surgery is pretty much never used anyway.

ghost:
Surgery for deconversion is really tedious and bad tbh

MacHac:
I changed the surgery because for debraining, it's
Scalpel - Retractors - Saw - Hemostat - Scalpel - Hemostat
and the old surgery, which did effectively the same thing, was 
Scalpel - Hemostat - Retractors - Saw - Scalpel

It just made no sense to me that there was randomly a hemostat in there for a procedure that otherwise looked so similar to debraining. (And, I feel, is implied to do something very similar).

@AdamElTablawy The surgery option is intended for special cases, or critical staff members.  For example, if the captain started seeing binary, the security staff could forcibly deconvert him if they captured him instead of shrugging their shoulders and taking a toolbox to his head.

Issue #1018, opened 2016-04-24
Whitelist Prefs Reset


ghost:
### Expected Behaviour

Your "Ultra" preference for AI stays Ultra.
### Actual Behaviour

I encountered this once, but I haven't tried to reproduce it, but the "Ultra" preference or whitelist preference for heads/AI resets after a round.

X-TheDark:
Duplicate of #752 
Nonetheless, don't close this. I'll take a look at this.

Issue #752, opened 2016-03-31
Ultra defaulting back to Never.


oisin100:
### Expected Behaviour

Stop going back to never and giving me annoying jobs.
### Actual Behaviour

Defaults back to never and gives me janitor. Cry in a corner.
### Steps to reproduce the problem
1. Set 'AI' preference to Ultra
2. Round restarts
3. AI preference is back to Never

If its any help im using a donor slot for my AI setup.

X-TheDark:
Make sure to hit the "Save Setup" button. This happens with all jobs resetting when your profile is reloaded if you didn't save the setup using the button.

oisin100:
Yep, I always do. It still resets.

X-TheDark:
Any additional information? Does it reset after the round-end thing?

oisin100:
I have no idea at what point it resets. It just seems to default back to never at the next round. Ill try again next round.

X-TheDark:
Yup, this does indeed, reset your preference from Ultra to Never. Damn, I'll see if I can fix this.

PR #1020, opened 2016-04-24
Fixed whitelist resetting

X-TheDark:
Refer to issue #1018 & #752
### Rookie mistakes

Who would have thought that the [src] of /datum/preferences/New(client/C) proc call is actually the datum itself, rather than the client? (Answer : Everyone but me and whomever (I know it's one of our guys) decided to make a check using [src] for is_donator)

Don't think changelog is necessary for this one.


oisin100:
Yey!

PR #1022, opened 2016-04-24
Adds magicarp and chaos carp to carp event

Time-Green:
85% for carp
5% chance for megacarp
5% for magicarp
2% for chaos magicarp

Thought more variation would be good.
This didn't take long to make, so if it's not good, no problem


Time-Green:
Due to weird reasons, the statue thing always comes with all my PR's.
I'll be fixing that now

Super3222:
This is pretty controversial.

Since no one should be turned into a random mob / teleported into the tesla / etc, on a round other than wizard. It just seems like a bad idea.

Time-Green:
The chance of it happening is rather small.
And it still happens from xenobiology

EMT321:
8 hours before choosing to implement magicarp and chaos carp to 7% of every round carp spawn?? Seems a bit rushed. 

I agree with super here, nobody should be getting random-insta murdered by NPC's.

PR #1024, opened 2016-04-24
Add command EVA suit

Time-Green:
Adds a command EVA suit.
-Is faster then most suits
-Moderate armor
-Kmc2000's sprite. 

Also comes with a new storage unit containing it
Spawns in the command EVA section
#221


Time-Green:
Great...
I messed up the map again.
5000 changes is not right.

Super3222:
Use mapmerge...

I can post how to do it correctly in #coder again if you want.

Time-Green:
I actually did. I am doing it over again as I made a mistake it seems

PR #1026, opened 2016-04-24
Refactored Species Code

Cruix:
Prelude to #975

Refactored species code with more hooks for possible future species changes, and moved some species code that was previously somewhat hacky into these hooks. This update by itself should cause no mechanical changes. I have tested everything I have changed or added, but I would appreciate others testing it as well.


TehFlaminTaco:
Do want.

Issue #993, opened 2016-04-22
Make lube more visible


EMT321:
### Expected Behaviour

Have a glimpse of a lubed surface, thus avoiding an 8 second stun which is impossible to avoid by walking, galoshes, or any non-traitor gear aside from active meth usage.
### Actual Behaviour

Lube is nigh-invisible, and there exists no counter to it other then crowbarring the tile. At which point it can be lubed again, and then there is no counter at all. While an outright counter is not what I am advocating for due to balance, I feel making it slightly more visible would alleviate some of the hallways of death without a station-based counter.

Heres an example of how subtle it currently is: Just a few pixels in a corner.: 
https://gyazo.com/720313fed48893885ee96968ac58807f

Much harder to spot when moving through an area, and not zoomed in: 
https://gyazo.com/042fc9811c3f249531d13e7758d49781
### Steps to reproduce the problem
1. Repeatedly lube an area around you.
2. Laugh at everyone running into it.
3. Pick between the helpless fully armored HoS, Captain, and janitor, all of whom are stunned regardless of equipment.

Groudonmaster:
Already nerfed enough it looks like water and magboots should counter them

ghost:
Giving them a shiny wet overlay, like /tg/ did water, would be useful.

Giving both water and lube the overlay, so you can't tell the difference between them.

Perhaps one or two pairs of security magboots with only 50% slowdown could spawn in the sec Suit Storage Units in the armory.

Super3222:
Lube was already nerfed to look like water. Maybe if you examine the tile, it should make the tile seem suspicious.

Super3222:
Lube was already nerfed to look like water. Maybe if you examine the tile, it should make the tile seems suspicious.

EMT321:
> Lube was already nerfed to look like water

I'm not really seeing this as a nerf.

Cruix:
It used to be totally invisible.

Super3222:
@EMT321 Eh, lube use to be completely invisible which was a nightmare when encountering a competent traitor chemist as Sec. /tg/ decided to make it look like water.

Seeing what Adam has said about a new overlay for both water and lube wouldn't hurt, I GUESS. An examine to differentiate them as well.

EMT321:
[...]

PR #1030, opened 2016-04-25
Makes wet/lubed floors SUPER obvious

X-TheDark:
Refer to issue #993.

Shared by both water/lube, so you can't really tell which is which.

wetness

If anyone wants to do a better overlay, feel free. I just upped the Alpha by about 3.5 times.

:cl: X-TheDark
tweak: Wet floors (Water/Lube) are now MUCH more visible.
/:cl:


EMT321:
Thank you, so very much.

Kmc2000:
RIP traitor janiborg

Groudonmaster:
Too hard of a counter :-1: x :100:

EMT321:
People can now see lube while moving != game breaking change. Lube has little to no counterplay. This still allows for lube in maintenance/darkened areas while not requiring you to shut the gravity off or wear magboots every time someone in medbay rolls traitor.

Earlier today on Rev a single lubed floor tile in the middle of the hall allowed enough time for the luber to come back and kill 4 people before they had a chance to stand up.

As to the objection "Just look harder, you suck at situational awareness": lube and water is only visible if you are actively slowly looking for it. I will call bullshit on people running at speed being able to see those dozen changed pixels on a single tile.

TGstation made it more visible, and cut down on lubepocalypse every round. Especially given how one can still spray maint and drag/push people onto lubed tiles, this will not make lube unusable. It will prevent it being a hard counter to the station, as it has been of late.

EMT321:
Plus, given the average level of awareness on the server, people are going to keep running into it. For reference: all janitor signs demarcating wet floors.

X-TheDark:
Honestly, maybe just adding a pair or two of security magboots like Adam suggested would be enough (they'd have slowdown, so you'd need to trade speed for the ability to not be slipped).

EMT321:
TGstation did what they did to prevent lube being a hard counter to the crew as a whole. That's why I wholly agree with making it more visible, so it comes down to player error rather then just having one of 3 roundstart items.

ghost:
merge it, this is not unbalanced

it works the same for both water AND lube

meaning you can see it instead of it being a tiny, impossible to spot dark tile

observant players will notice, not others

lube and water share the same sprite, so if you try to walk on water OOPS ITS LUBE and you go sailing

DavidEdwards:
Lets give it a try! I'm a little bit skeptical about making it more dark, but we can always revert.

Issue #1031, opened 2016-04-25
Borer Paralysis causing infect to autofail.


Stealthkibbler:
### Expected Behaviour

Target gets stunned by psychic lance, borer climbs into ear with no resistance
### Actual Behaviour

Borer constantly gets dislodged despite host paralysed and not moving at all.
### Steps to reproduce the problem
1. Position yourself next to someone
2. Hit them with psychic lance
3. Attempt to infect but get the error message saying that the host is moving around.

I think the problem is the victim is spamming their move keys so the infect registers that as them moving despite their sprite unable to move from the psychic paralyze.

Wasnt just me, other borer said it happened to him too.

X-TheDark:
Yup, just tested this. They don't even have to move, it automatically fails during the paralysis. I can fix this.

Stealthkibbler:
Changed title, Also it works when you normal infect without paralysing them.

PR #1032, opened 2016-04-25
Borers do not autofail infection during paralysis

X-TheDark:
Refer to issue #1031 .

Borers can infect during paralysis and it also shows the progress bar now!

:cl: X-TheDark
bugfix: Borers can now actually infect during paralysis.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_powers.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_powers.dm
@@ -25,7 +25,7 @@
 		src << "<span class='warning'>You slither up [H] and begin probing at their ear canal...</span>"
 		src.layer = MOB_LAYER
 		//H << "<span class='userdanger'>You feel something trying to enter your ear...</span>"
-		if(!do_after(H,30))
+		if(!do_after(src, 30, 5, 0, H))
 			src << "<span class='warning'>As [H] moves away, you are dislodged and fall to the ground.</span>"
 			return
 


code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -25,7 +25,7 @@
 		src << "<span class='warning'>You slither up [H] and begin probing at their ear canal...</span>"
 		src.layer = MOB_LAYER
 		//H << "<span class='userdanger'>You feel something trying to enter your ear...</span>"
-		if(!do_after(H,30))
+		if(!do_after(src, 30, 5, 0, H))
```
X-TheDark:
```
/proc/do_after(mob/user, delay, numticks = 5, needhand = 1, atom/target = null)
```
Rest defined as numbers because not using named arguments is faster than using them.

ghost:
remove the 'GET THE FUCK AWAY SOMETHING IS CRAWLING INTO YOUR EAR' message

it makes stealth infection impossible and your host will IMMEDIATELY run to medbay for sugar before you can get any chems to punish/stop them

X-TheDark:
EDIT: There's actually no message when the borer is trying to (or fails to) infect you. You only get the message when it uses Paralyze. My bad.

Like that message even makes a difference...the entire pair of "YOU ARE FROZEN/SOMETHING IS CRAWLING UP YOUR EAR" does not allow you to stealth. Nor that it should.

It anything, random chem system for pacifying the borer should be added. And the chem should be given out only when you reach like 100% influence (You suddenly crave [Chemical X]).

ShadowDeath6:
If there's a message that shows up when the borer tries to infect you (not paralyze) then that's my fault.

Kmc2000:
You feel something brush against your feet.... 

That work? pretty generic

(As it slithers up in yo ear hole)

X-TheDark:
That wouldn't work. Any message that can be uniquely identified as belonging to borer infecting you will give them away. If anything, there shouldn't be ANY messages of you being actually infected (I mean, not like it isn't obvious with a brain slug right next to you, unless it's under the table or anything), because I think (correct me if I'm wrong), you actually get some messages about something being in your head when you get enough influence.

ghost:
borers should be able to stealthily infect if they don't paralyze their host.

that's their intent.

if you're hiding up a bar table and crawl into someone's ear you should be unnoticeable until they move away, like pickpocketing, so that you can actually stealth infect. otherwise you get into your host with too low chems to defend yourself and too low influence to defend yourself and they immediately rush for the chef's table and guzzle sugar.

AsV9:
@AdamElTablawy 

> remove the 'GET THE FUCK AWAY SOMETHING IS CRAWLING INTO YOUR EAR' message

I don't understand. It's commented out. There shouldn't be a message..

There is no `H << "Stuff"` anywhere if I am reading this correctly.

X-TheDark:
Oh, damn. That's actually so weird, I could have sworn that I saw a message telling me something is trying to crawl into the ear thing.

There actually is NO message, if you just try to infect. The only time you get the message is actually when you use Paralyze (not even failing the infection because target moved will give it a message)...my bad.

ShadowDeath6:
Can we merge this

DavidEdwards:
Indeed

PR #1050, opened 2016-04-27
Fixed bugs with block chance

### LINKED ISSUE #1040: Nerf double-bladed energy sword, allow batons the ability to knock down.
### After one of our recent updates, double eswords now block literally everything. That means melee; stunbatons, telescopic batons, anything besides a bottle to the head, which doesn't really work on nukeops.

12 TC turns you into a walking tank. After witnessing a single near-dead nukeop below orange take down an entire squad of ERT by just holding a sword, I think this is an OBVIOUS problem.

There's literally no reason to buy anything other than a double esword as a nukeop, as one or even two people could steamroll the entire station, simply because they can't be touched. That is insane. There's no chance, at least before you could use a baton and bring them down, but now the only real choice is to counter with space lube. That's a joke.

Remove the blockage of energy swords from deflecting batons. Allow users of energy swords to be knocked down from their stuns, aka, revert eswords back to the way they were. I don't possible see how this can stay a thing.

Cruix:
Fixes #1040 

:cl:
bugfix: dual eswords no longer have 100% block chance.
bugfix: blocking attacks with shields or blocking weapons now properly shows a message that the attack was blocked.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -5,7 +5,7 @@
 /obj/item/weapon/shield/hit_reaction(mob/living/carbon/human/owner, attack_text, final_block_chance, damage, attack_type)
 	if(attack_type == THROWN_PROJECTILE_ATTACK)
 		final_block_chance += 30
-	return ..()
+	return ..(owner, attack_text, final_block_chance, damage, attack_type)
 
 /obj/item/weapon/shield/riot
 	name = "riot shield"
--- a/code/datums/martial.dm
+++ b/code/datums/martial.dm
@@ -550,7 +550,6 @@
 	return ..()
 
 /obj/item/weapon/twohanded/bostaff/hit_reaction(mob/living/carbon/human/owner, attack_text, final_block_chance)
-	if(wielded && prob(final_block_chance))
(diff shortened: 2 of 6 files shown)


ShadowDeath6:
tbh I'd like for bostaffs to have 100% melee block, it would be a cool buff to the shitty gang

code/game/objects/items/weapons/shields.dm
```diff
@@ -5,7 +5,7 @@
 /obj/item/weapon/shield/hit_reaction(mob/living/carbon/human/owner, attack_text, final_block_chance, damage, attack_type)
 	if(attack_type == THROWN_PROJECTILE_ATTACK)
 		final_block_chance += 30
-	return ..()
+	return ..(owner, attack_text, final_block_chance, damage, attack_type)
```
X-TheDark, line 8: Why? Doesn't ..() automatically pass the same parameters it gets?

Cruix, line 8: It does, but on the line above one of the parameters is changed, and if I understand correctly the change will not be sent to the ..() call.

To be honest I could not test this because every single shield I found in the code overrides this proc.

PR #1061, opened 2016-04-30
First batch of runtime fixes and code refactoring.

Nirnael:
### Runtime fixes from runtime log of 22-04-2016

More fixes to come but I want to update the runtime log and the code to the most recent.
Messy fix description because too lazy:
-breathe runtime fixed
-thrownby runtime fixed
-cybermen images runtime fixed
-cybermen hallucination cut runtime fixed
-corgi update fluff runtime fixed
-storage show to proc client runtime fixed
-hallucination xeno runtime fixed
-reagents null runtime fixed

:cl:
bugfix: First batch of runtime fixes and code refactoring of 22-04-2016 runtime log.
/:cl:


code/modules/flufftext/Hallucination.dm
```diff
@@ -185,22 +185,22 @@ Gunshots/explosions/opening doors/less rare audio (done)
 /obj/effect/hallucination/xeno_attack/New(loc,var/mob/living/carbon/T)
 	target = T
 	for(var/obj/machinery/atmospherics/components/unary/vent_pump/U in orange(7,target))
-		if(!U.welded)
+		if(U.welded)
```
oisin100: This makes no sense, Xeno hallucination only works on welded vents? Dead giveaway that its a hallucination.
Also, With this code all welded vents in a range of 7 from the target will have their own xeno hallucinations.

code/game/gamemodes/cybermen/cybermen_hacking.dm
```diff
@@ -1110,11 +1110,14 @@
 
 /obj/effect/hallucination/cybermen_conversion/New(loc, var/mob/living/carbon/T)
 	target = T
+	matrix_turfs = new/list()
+	matrix_images = new/list()
```
Cruix: Is this necessary? Both of these vars are initialized as lists.

code/game/gamemodes/cybermen/cybermen_hacking.dm
```diff
@@ -1110,11 +1110,14 @@
...
 	..()
 
 
 /obj/effect/hallucination/cybermen_conversion/process()
-	target.client.images.Remove(matrix_images)
+	if(target.client)
```
Cruix: May as well just make it return if there is no client, otherwise it will just runtime below.

code/game/gamemodes/cybermen/cybermen_objectives.dm
```diff
@@ -171,8 +171,7 @@
 
 /datum/objective/cybermen/expand/hack_ai/New()
 	..()
-	for(var/V in ai_list)
-		var/mob/living/silicon/ai/new_ai = V
+	for(var/mob/living/silicon/ai/new_ai in ai_list)
```
Cruix: This was intended as an istypeless forloop. Unless there is ever a situation where a non-AI is in ai_list, which there should never be, this has worse performance.

code/modules/mob/living/simple_animal/friendly/dog.dm
```diff
@@ -275,7 +274,7 @@
 	return valid
 
 /mob/living/simple_animal/pet/dog/corgi/proc/update_corgi_fluff()
-	switch(src.inventory_head.type)
+	switch(inventory_head.type)
```
Cruix, line 282: src should never be null in this situation, did you find some situation where it was?

DavidEdwards:
Some of my merges today must have affected your PR. Could you check the conflicts, and also check Cruixs comments.

code/modules/flufftext/Hallucination.dm
Nirnael: > This makes no sense, Xeno hallucination only works on welded vents? Dead giveaway that its a hallucination.
> Also, With this code all welded vents in a range of 7 from the target will have their own xeno hallucinations.

Hallucinations never work otherwise because all the pumps are welded, and no it only runs once because it breaks out of the for.

code/game/gamemodes/cybermen/cybermen_hacking.dm
Nirnael: Nope, missed it being initialized, thanks.

code/game/gamemodes/cybermen/cybermen_hacking.dm
Nirnael: Yeah, thanks.

code/game/gamemodes/cybermen/cybermen_objectives.dm
Nirnael: How does it have worse performance?

code/modules/mob/living/simple_animal/friendly/dog.dm
Nirnael, line 282: No, just removing src because its unnecessary.

code/modules/flufftext/Hallucination.dm
oisin100: Ah, Sorry. Did not see the break statement.

You want the hallucination to only work on Welded vents?

code/game/gamemodes/cybermen/cybermen_objectives.dm
Cruix: It calls istype() on every element in the list, which is not necessary because we know that everything in the list is a /mob/living/silicon/ai

code/modules/flufftext/Hallucination.dm
Nirnael: Yeah otherwise it never works because the only place you could find unwelded vents is from bombs having exploded and blown the vents apart.

code/game/gamemodes/cybermen/cybermen_objectives.dm
Nirnael: Indeed it is, thanks.

code/modules/mob/living/simple_animal/friendly/dog.dm
Cruix, line 282: :+1: It baffles me when people use src.variable for no reason.

code/modules/flufftext/Hallucination.dm
oisin100: Vents start unwelded. I think that what you're thinking of is `unsecured` vents which are vents that have been unwrenched by a wrench, `welded` vents are vents that have been welded with a welding tool.

Nirnael: Ah right, i thought you had to weld them first, thanks.

code/game/gamemodes/cybermen/cybermen_hacking.dm
```diff
@@ -1130,8 +1131,7 @@
 		for(var/turf/T in matrix_turfs)
 			matrix_images += image(image_icon,T,image_state,MOB_LAYER)
 
-		if(target.client)
-			target.client.images |= matrix_images
+		target.client.images |= matrix_images
```
oisin100, line 1134: What if the client is somehow null? This is just gonna create more runtimes.

Cruix, line 1134: It would have already runtimed on line 1118

oisin100, line 1134: Ah yes.

PR #1062, opened 2016-04-30
Replace the runtime condenser.

george99g:

Replacing /tg/'s runtime condenser with one that plays more easily with scripts.
Also making the code's formatting a little bit more readable.


MrsTonedOne:
just to note, i recently added more improvements to this that you should also port.

https://github.com/tgstation/-tg-station/pull/17440

also, i might port your command line extensions, with the exception i'd make it default to old behavior so that you can drag and drop a file onto the exe to have it process it. and support stdin to stdout for piping a file to it. I'd keep the pause code thou, if you could make it detect if it was ran standalone or via a terminal so it can be ran by users and still display any errors, but work for scripts.

On linux at least, it works on scripts because they don't have a tty so the char read returns immediately.

george99g:
> just to note, i recently added more improvements to this that you should also port.

Alright. I'll take a look at porting it over when I have the time.

> and support stdin to stdout for piping a file to it. I'd keep the pause code thou, if you could make it detect if it was ran standalone or via a terminal so it can be ran by users and still display any errors, but work for scripts.

Detecting the pipes without io.h or digging yourself into a bunch of library dependencies will be a pain to do.

> On linux at least, it works on scripts because they don't have a tty so the char read returns immediately.

I haven't actually noticed that. I'm used to command line executables returning as soon as possible because you can just keep the terminal open. It would probably be a pain if you had to press an extra return to get it away so you can start typing commands again.

Then again, one could probably write an OS check to add it for Windows but not Linux.

Issue #1074, opened 2016-05-02
Borgs are exploitable


ToGWtF:
> [7:16 AM] nirnael: well
> [7:16 AM] nirnael: when you are standard
> [7:16 AM] nirnael: and click to select module
> [7:16 AM] nirnael: click a few more times
> [7:16 AM] nirnael: then select borg
> [7:16 AM] nirnael: and it will open another selection window
> [7:16 AM] nirnael: as many as you clicked
> [7:16 AM] nirnael: so you can infintely change borgs
> [7:17 AM] nirnael: and keep stuff

tested and reproduced

PR #1075, opened 2016-05-02
Borgs are NOT exploitable

X-TheDark:
Refer to issue #1074.

Prevents borgs from using multiple open Select Module windows to change modules on the fly, while keeping all the stuff in the hotbar. Changes you into the first selected module, everything else is disregarded (even if you can still click on stuff).

I am not sure if there's a better way to do this other than endless returns everywhere. I really don't want to define a new variable in_use or something in borgs just for this.

EDIT: Thanks to Nirnael (@fizzle7) for pointing me into the right direction and avoiding all the returns.


ShadowDeath6:
Fun removal xddd

PR #1107, opened 2016-05-07
Fixes some mindslave bugs with real_name

oisin100:
Game log logged the players name and not their Ckey.
Mindslave implant used the `name` variable instead of `real_name` meaning that you could wear someone elses ID to confuse the implant.

:cl:
bugfix: Mindslave implant is no longer fooled by wearing someone elses ID.
/:cl:


ShadowDeath6:
Can you make golem runes work like this as well

oisin100:
> Can you make golem runes work like this as well

Ill take a look.

PR #1141, opened 2016-05-12
Rules broken are now added to ban reason

AsV9:
Adds a new text prompt for when bans are added which specifies which rule a player has broken.

image

Image above is nearly respective of what it looks like ingame.

[https://forums.yogstation.net/index.php?threads/vote-should-ban-reasons-have-which-rules-bans-are-based-on-stated-in-them.10454/](https://forums.yogstation.net/index.php?threads/vote-should-ban-reasons-have-which-rules-bans-are-based-on-stated-in-them.10454/)

:cl:AsV9
tweak: Slight change to bans, ban reasons will now have a rule attached to them for reference in ban appeals. 
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -919,6 +919,7 @@
 					var/rules_broken = input(usr,"Which rule did they break? (Rule broken number.)",,"0.0") as text|null
 					if(!rules_broken)
 						return
+					reason += " ([rules_broken])"
 
 					var/msg
 					for(var/job in notbannedlist)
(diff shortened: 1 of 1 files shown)


ShadowDeath6:
Nerf ban reasons when :+1:

Super3222:
This is a nice change. :+1:

ghost:
What if no specific rules can be cited in the ban but they did something banworthy anyway? That happens, where someone does something very banworthy but not specifically mentioned in the rules.

AsV9:
Then type that then. It's `as text|null` for a raisin.

george99g:
> What if no specific rules can be cited in the ban but they did something banworthy anyway? That happens, where someone does something very banworthy but not specifically mentioned in the rules.

There's always 0.0 if they're being dicks and none of the rules cover it.

AsV9:
The input field defaults to 0.0 so.. not really an issue.

oisin100:
I suppose its worth a try.

PR #1159, opened 2016-05-14
Chat rework and chat tabs.

Nirnael:
**EDIT2: Apparently goon has javscript in their terminal for fancy chat, so I'm gonna close this, learn JS and use that to make better tabs and chat, which probably won't be any soon so this has to be closed.**
EDIT: I'm fully reworking chat from the ground up, fuck this, I'll try to make it as customizable as possible.
Pre-PR to see what people want and think. Will post images and changes soon.

Basically this, if you want anything else, suggest it.

### DIFF
--- a/code/game/verbs/ooc.dm
+++ b/code/game/verbs/ooc.dm
@@ -1,7 +1,9 @@
 /client/var/bypass_ooc_approval = 0
-/client/verb/ooc(msg as text)
-	set name = "OOC" //Gave this shit a shorter name so you only have to time out "ooc" rather than "ooc message" to use it --NeoFite
-	set category = "OOC"
+/client/verb/ooc(msg as text) //called by callOOC(callOOC appears in the verb list, not this one) or other normal calls
+	set hidden = 1
+	//saved for legacy purposes and contribution commentary, remove if you like
+	//set name = "OOC" //Gave this shit a shorter name so you only have to time out "ooc" rather than "ooc message" to use it --NeoFite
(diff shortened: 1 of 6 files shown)


AsV9:
I'd like a mode to merge everything like the current system because I prefer oversight without swapping over modes all the time. But I can see why people would want to filter it.

oisin100:
> IC, OOC, Admin log.

IC And OOC and Asay should be kept together

Admin Log should get its own Tab

> Preferences to keep old system or have the new tab system.

Yes, I like the idea(And i'm sure people will prefer to use the new system more than the old) but i like all  my chat in the same place.

> Any output should go in 2 tabs at the same time, or more? Round restart will definitely go into IC and OOC for ease.

If both IC and OOC are in the same tab this doesn't matter

> Finish redirecting all output to the appropriate tab.

Good lord, Good luck with that!

Nirnael:
The output actually is easy, the good thing about OOC and admin log is it's all in a single proc.
I will add preferences then, possibly even for OOC and asay to choose a tab.

ShadowDeath6:
I'd like an option to have all the stuff in the same place as it is now

Nirnael:
Sigh, in the end all this is gonna be for nothing but at least then I'll be able to retire with a reason.

X-TheDark:
I think it'd be pretty damn useful to have separate channels like this.

ShadowDeath6:
Rip

PR #1166, opened 2016-05-15
Remove BYGEX information from the readme.

george99g:
With regex handled by BYOND directly, do we really need information about the old system in the readme?


AsV9:
why didn't you just commit directly to dev

george99g:
> why didn't you just commit directly to dev

Because raisins.
And someone probably not being okay with this change somehowl.

AsV9:
that's stupid. you're a `council member` for a reason, don't be so hesitant

george99g:
> that's stupid. you're a council member for a reason, don't be so hesitant

I should have done a vote on it instead, then.
Also, I can't believe I actually mistyped somehow.

PR #1182, opened 2016-05-18
Ports fast process.

oisin100:

Ports Fastprocess and Super fun conveyor belts.
#### Changelog

:cl:
rscadd: Ported fast process from /TG/, Credit goes to @AndrewJacksonThe2nd
tweak: Conveyor belts no longer take forever to move stuff, infact they move items so quickly you can barely get a glimpse of them.
tweak: Ore redemption machines no longer take FOREVER to eat up ores.
/:cl:

I was tempted to commit my change for the singularity fast process... But that just turned the singularity into a Overpowered machine of pure destruction.

### DIFF
--- a/code/game/machinery/machinery.dm
+++ b/code/game/machinery/machinery.dm
@@ -118,18 +118,25 @@ Class Procs:
 	var/interact_offline = 0 // Can the machine be interacted with while de-powered.
 	var/mob/living/silicon/pai/paired
 	var/paiallowed = 0
+	var/fast_process = 0
 
 /obj/machinery/New()
 	..()
 	machines += src
-	SSmachine.processing += src
+	if(fast_process)
+		SSmachine.processing += src
+	else
+		SSfastprocess.processing += src
 	power_change()
 
 /obj/machinery/Destroy()
 	if(paired)
 		paired.unpair(0)
 	machines.Remove(src)
-	SSmachine.processing -= src
+	if(fast_process)
+		SSfastprocess.processing = src
+	else
+		SSmachine.processing -= src
 	if(occupant)
 		dropContents()
 	..()
--- a/code/controllers/subsystem/fastprocess.dm
+++ b/code/controllers/subsystem/fastprocess.dm
@@ -1,0 +1,37 @@
+var/datum/subsystem/fastprocess/SSfastprocess
+
+/datum/subsystem/fastprocess
+	name = "Fast Process"
+	priority = 12
+	wait = 1
+	dynamic_wait = 1
+	dwait_upper = 10
+	dwait_buffer = 0
+	dwait_lower = 1
+	dwait_delta = 6
+
+	var/list/processing = list()
+	var/list/currentrun = list()
+
+/datum/subsystem/fastprocess/New()
+	NEW_SS_GLOBAL(SSfastprocess)
+
+/datum/subsystem/fastprocess/stat_entry()
+	..("FP:[processing.len]")
+
+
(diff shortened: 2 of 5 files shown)


ShadowDeath6:
isn't that goofball

ghost:
make sure that you port the foam fix too because this will fuck foam

oisin100:
> make sure that you port the foam fix too because this will fuck foam

I did not port the FastProcess Foam(Thus it should not be borked), I like foam how it is.

oisin100:
Heres a GIF of the conveyor belts, http://imgur.com/3DsnBq7

code/game/machinery/machinery.dm
```diff
@@ -118,18 +118,25 @@ Class Procs:
...
+	var/fast_process = 0
 
 /obj/machinery/New()
 	..()
 	machines += src
-	SSmachine.processing += src
+	if(fast_process)
+		SSmachine.processing += src
+	else
+		SSfastprocess.processing += src
```
X-TheDark, line 129: Other way around.

X-TheDark:
I think this is too fast, honestly, maybe just set the wait to 10? (So belts are processed once every second instead of once every 2)

ghost:
No, this is good

X-TheDark:
While I don't mind making belts a bit faster, there must be SOME chance to ctrl+click someone or pick up an item without it whoosing by you at mach LUDICROUS.

code/game/machinery/machinery.dm
oisin100, line 129: Woops!

oisin100:
> I think this is too fast, honestly, maybe just set the wait to 10? (So belts are processed once every second instead of once every 2)

Ill see what i can do to slow it down.

MrsTonedOne:
Fair warning, @AndrewJacksonThe2nd is iamgoofball.

Also, resist any temptation to add more things to this. it can easily get clogged down and slow back down.

X-TheDark:
Thanks for the warning, MSO.

Also, all you need to do to slow this down is to change `wait = 1` to `wait = 10` to slow it down to the extent I described.

ghost:
NO.

I DO NOT WANT THIS SLOWED DOWN.

Conveyor belts are almost NEVER used for anything.

Just keep it at the current rate and merge it. 

@oisin100 

It's current speed is perfect and conveyor belts are never used so what's the harm?

MrsTonedOne:
also, changing wait has no effect, this is a dynamic wait subsystem, so after every run, the mc dynamically reschedules it based on current load of the subsystem, the mc, and the game as a whole.

Iamgoofball:
@oisin100 
@X-TheDark 

Hi, if you're trying to slow down fastprocess you don't understand how it works.

The thing about fastprocess is its literally trying to process as fast as possible WITHOUT causing lag.

Slowing down isn't possible without disabling the dynamic wait system, and is literally killing the point of it

PS. port the foam speedups too, the whole point of fastprocess was actually to speed up foam processing

Iamgoofball:
> I was tempted to commit my change for the singularity fast process... But that just turned the singularity into a Overpowered machine of pure destruction.

we did some runs of this on Sybil, even with the lagfixes it gets excessive

oisin100:
> Slowing down isn't possible without disabling the dynamic wait system, and is literally killing the point of it

You could use a cooldown(Not using Sleep god damn), and then mark objects as moved already so the next conveyor doesn't move the same object in the same iteration. I just haven't had the time to do it.

Anyways, closing this now, it will be done in the rebase.

Issue #34, opened 2016-01-16
Memo system not working


george99g:
opened by Alex West · 2 months ago
Can't write/view/edit memos in any way, and any(if they are still set from before we moved host) that do exist aren't showing up when people join.
The menu comes up and then none of the buttons appear to do anything.

PR #1189, opened 2016-05-20
Fixes memo system to not work on a database system.

AsV9:
Fixes issue #34 

Fixes the memo system by returning to the savefile method for saving it. Frankly having the memo system work on a database basis is unnecessary and it's confusing with the DB spaghetti code involved. It's not worked for a solid 8 months and it's about time we get it back.
#### Changelog

:cl:
bugfix: Fixes the admin memo system.
/:cl:


oisin100:
"Fixes" = "To not work"

Fixing something so it doesent work? 10/10.

PR #1195, opened 2016-05-21
Blub Balance PR 2

ghost:

Borrows /tg/'s system of decreased blob expansion at farther ranges.
#### Changelog

:cl:
rscadd: Blob automatic expansion chance decreases as the blob gets farther away from its core.
/:cl:

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1776,3 +1776,23 @@ B --><-- A
 				closest_distance = distance
 				closest_atom = A
 	return closest_atom
+
+//ultra range (no limitations on distance, faster than range for distances > 8); including areas drastically decreases performance
+// stole- *cough* ported from /tg/
+/proc/urange(dist=0, atom/center=usr, orange=0, areas=0)
+	if(!dist)
+		if(!orange)
+			return list(center)
+		else
+			return list()
+
+	var/list/turfs = RANGE_TURFS(dist, center)
+	if(orange)
+		turfs -= get_turf(center)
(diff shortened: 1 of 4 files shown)


FluffySurvivor:
make the biohazard announcement appear in the place of the "classified" announcement so admins cant yell meta. 

helps so people can get fighting faster too

ghost:
I don't know about that, Fluffy.

ShadowDeath6:
It's not meta, and even if it is you can wait like 3 minutes for it anyway

ShadowDeath6:
It's not meta, and even if it is you can wait like 3 minutes for it anyway

ghost:
nobody gives a fuck if you yell BLBO because of classified

but yeah sure make it show in place of the classified

ALSO

it's worth mentioning that past a certain point, blob really can't be stopped anyway

making it weaker to fire, welders, lasers, or something might be nice

or maybe just making killing a node disable blob point regen temporarily since it usually has a metric fuckton of them

FluffySurvivor:
i really don't want people to suffer because of some metahitler admin

ShadowDeath6:
Some blob types actually die in one laser shot anyway

ghost:
it'd be good to make ALL blob types die in one laser shot so that lasers aren't completely useless versus a big blob

buff fire maybe too to kill normal blob tiles and nodes much faster

oisin100:
> Blob automatic expansion chance decreases at rangers farther from the core.

Blob automatic expansion chance decreases as the blob gets farther away from the core.

Issue #1219, opened 2016-05-27
Fireball stops recharging if you are knocked into crit while holding it


Cruix:
### Expected Behaviour

When knocked into crit while holding fireball, it drops and explodes just like if you had rested.
### Actual Behaviour

It disappears and you never get the cooldown back.
### Steps to reproduce the problem
1. Give yourself fireball.
2. Cast it but do not throw it.
3. deal 101 brute damage to yourself, then heal yourself, and be unable to ever cast it again.

PR #1227, opened 2016-05-28
Fixed fireball not working when critted while holding it

X-TheDark:
Refer to issue #1219 .

One of the most peculiar and interesting bugs I had to fix.

**Cause:** Basically, if we dropped the fireball due to crit-damage, **somewhere down the proc call stack _usr_ (BYOND built-in var) started returning null**, which caused line 83 `var/obj/effect/fire_ball/F = new(get_turf(usr)) (before fix)` to create the fireball (the thing that gets created when you drop/throw the thing you hold in hand) in null-space, fucking everything up. This also applied to invocation(), that also needed to have user passed explicitly, rather than using usr.

One of the best showcases of the big bold text from http://www.byond.com/docs/ref/info.html#/proc/var/usr

**Fix:** Explicitly pass the user to the proc, instead of using usr inside it. Fixes both fireball not triggering and it not being able to recharge ever.
#### Changelog

:cl: X-TheDark
bugfix: Fireball now correctly triggers when the holder drops it due to crit-damage.
/:cl:

### DIFF
--- a/code/datums/spell.dm
+++ b/code/datums/spell.dm
@@ -170,15 +170,16 @@ var/list/spells = typesof(/obj/effect/proc_holder/spell) //needed for the badmin
 
 /obj/effect/proc_holder/spell/proc/perform(list/targets, recharge = 1, mob/user = usr) //if recharge is started is important for the trigger spells
 	before_cast(targets)
-	invocation()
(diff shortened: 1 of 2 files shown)


oisin100:
Interesting...

Issue #1226, opened 2016-05-27
Revolver


Keekenox:
40 damage, 20 stamina

ghost:
eh, i've always hated stamina damage as a mechanic-it'll slow the guy just as much

i'd prefer 40 damage and a lower clip

PR #1228, opened 2016-05-28
Nerfs revolver damage to 35

X-TheDark:
Refer to issue #1226.

Basically, the intent is so you cannot kill as many people with the revolver with 1 clip.

Changes the ammo type of the revolver cylinder to heavybullet, that does 35 damage.
So now, instead of being able to crit people in 2 shots, you need 3 (for unarmoured people).

Boolet points:
- Shooting HoS in the chest with this 4 times will deal 98 damage.
- Total damage went down from 420 to 245
- Crits in 3 instead of 2
#### Changelog

:cl: X-TheDark
rscadd: Revolver does 35 damage per shot instead of 60.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/ammo_casings.dm
+++ b/code/modules/projectiles/ammunition/ammo_casings.dm
@@ -1,7 +1,7 @@
 /obj/item/ammo_casing/a357
 	desc = "A .357 bullet casing."
 	caliber = "357"
-	projectile_type = /obj/item/projectile/bullet
+	projectile_type = /obj/item/projectile/bullet/heavybullet
 
 /obj/item/ammo_casing/a50
 	desc = "A .50AE bullet casing."


FluffySurvivor:
This is a pretty heavy nerf and the revolver will lose its niche and become... A weaker stetchkin.

X-TheDark:
Hmmm. That's true, the pistol does 30 damage per shot. This would make the revolver pretty much obsolete (while costing more).

But what would the revolver's niche even be? I certainly agree that critting in 2 hits is a bit...much. And being able to kill 3 people with 1 magazine is not exactly what I would call a niche, it's just a really, really, really strong handgun.

If it's supposed to be a weapon for either quick kills or defense, I could leave it with 60 damage, but have it only contain 2 shots, so you can kill one person easily, but nothing more. Would be accompanied by some price decrease for revolver.

Alternatively, this could be closed and people can be told to "get robust".

ghost:
100% armor pen?

FluffySurvivor:
up uplink ammo cost to 4, make recycled speedloaders only give half of the metal back.
damage down to 50

The point here is to make it less effective against security and ammo being less accesible. This way you will preserve the niche of "risky but high firepower".

X-TheDark:
I'll close this one for now. Maybe I'll redo it later with the whole autolathe half resource for speedloader thing, since it's a rather big feature to have items redeemable at a different rate than 1:1.

PR #1278, opened 2016-06-14
Ports /tg/ random View Variables crash fix

X-TheDark:

Sometimes, View Variables seems to crash the client. Hopefully, this fixes that.
Credit to @lordpidey and his PR:https://github.com/tgstation/tgstation/pull/18334

### DIFF
--- a/code/datums/datumvars.dm
+++ b/code/datums/datumvars.dm
@@ -288,7 +288,7 @@
 	var/list/names = list()
 	for (var/V in D.vars)
 		names += V
-	sleep(1)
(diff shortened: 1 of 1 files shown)


lordpidey:
Might not be a bad idea to add a comment about why, so noone removes the "useless" sleep.

X-TheDark:
Good idea, I thought it wasn't necessary, but I guess it's better safe than sorry. This will only be an issue until we finish the rebase (as this is included there). Consider this a temporary fix.

oisin100:
Outdone yourself with this port Dark.

